PR #24, opened 2022-08-17
Fixed timesteps followed  by log timesteps method. Added options for …

fionntancallan:
…different timestep set ups in artisoptions


artisoptions_nltenebular.h
```diff
@@ -215,4 +215,22 @@ static bool WRITE_PARTIAL_EMISSIONABSORPTIONSPEC = false;
...
+// values that would give the same number or more more fixed timesteps than the total number of timesteps in the
+// simulation. The times are set in days.
+
+static const bool LOG_TIMESTEPS = true;
+
+static const bool FIXED_TIMESTEPS = false;
+
+static const bool LOG_THEN_FIXED_TIMESTEPS = false;
+
+static const bool FIXED_THEN_LOG_TIMESTEPS = false;
```
lukeshingles, line 230: Of the 16 ways to set these four booleans, 12 of them are invalid and I don't see any lines to detect this and present an error message. Replace four switches with an enumeration like this:

```
enum timestepsizemethods {
  TIMESTEP_SIZES_LOGARITHMIC = 0,
  TIMESTEP_SIZES_CONSTANT = 1,
  TIMESTEP_SIZES_CONSTANT_THEN_LOGARITHMIC = 2,
  TIMESTEP_SIZES_LOGARITHMIC_THEN_CONSTANT = 3,
};

static const enum timestepsizemethods TIMESTEP_SIZE_METHOD = TIMESTEP_SIZES_LOGARITHMIC;

```

input.cc
```diff
@@ -2378,53 +2378,101 @@ void time_init(void)
   globals::time_step = (struct time *) malloc((globals::ntstep + 1) * sizeof(struct time));
 
   /// Now set the individual time steps
-  for (int n = 0; n < globals::ntstep; n++)
+  if (LOG_TIMESTEPS)
```
lukeshingles, line 2381: Update to handle the enumeration in artisoptions.h. Maybe use a switch statement?

lukeshingles (changes requested):
The automated test is failing because you need to set the new options in tests/nebularonezone_inputfiles/artisoptions.h

lukeshingles (approved):

PR #29, opened 2022-11-15
2022 November release

lukeshingles:
- For kilonovae, enable seeding with initial energy in the q column of model.txt
- Add kilonova test calculation to GitHub actions (to test alpha/beta decays and thermalisation)
- exspec: read all packets files into memory if total size is below a limit (maxpktmem_mb = 6000). This makes angle-resolved files much faster to generate.
- Prepend each log line with the date and time
- Fix escaping direction bin for non-normalised direction vectors
- Reduce memory fragmentation by making block allocation for atomic transition arrays and sharing on node
- Use C++ std library search functions (faster and less code)

A 1D Sim+2010 Sub-MCh test with classic options gives about an 8-11% decrease in run time.


AlexHls (changes requested):
Code does not compile with GCC 11, requires GCC12 which might not be available on compute clusters

lukeshingles:
> Code does not compile with GCC 11, requires GCC12 which might not be available on compute clusters

Thanks. I've fixed compilation with GCC 11.2 (on JUWELS).

AlexHls:
Yes it does indeed fix it if one uses the OpenMPI mpicxx. For the Intel compiler (which is loaded by the job scripts), it does still fail on JUWELS. Is there any specific reason for using the Intel compilers in the job scripts? Otherwise we should consider changing this in the job scripts to match the github workflow.

AlexHls:
Yes, you are right, I've missed that change. My bad.

lukeshingles:
It should compile with the Intel Compiler 2021 now (but I haven't actually run the binary).

PR #34, opened 2023-05-09
Read 2 phixs files for hybrid atomic data simulations

fionntancallan:
Added switch to allow choice of reading one or two phixsdata files depending on the atomic data configuration being used


input.cc
```diff
@@ -233,25 +233,33 @@ static void read_phixs_data_table(FILE *phixsdata, const int nphixspoints_inputt
 }
 
 static void read_phixs_data(int phixs_file_version) {
-  globals::nbfcontinua_ground = 0;
-  globals::nbfcontinua = 0;
+  if (USE_TWO_PHIXS_FILES == false) {
+    globals::nbfcontinua_ground = 0;
+    globals::nbfcontinua = 0;
+  }
```
lukeshingles, line 239: Is this special case needed? It looks like you could set these to zero in read_atomicdata_files() just as you do when USE_TWO_PHIXS_FILES == true.

input.cc
```diff
@@ -233,18 +233,26 @@ static void read_phixs_data_table(FILE *phixsdata, const int nphixspoints_inputt
...
+        "phixsdata.txt data\n",
+        globals::NPHIXSPOINTS, globals::NPHIXSNUINCREMENT);
+    last_phixs_nuovernuedge = (1.0 + globals::NPHIXSNUINCREMENT * (globals::NPHIXSPOINTS - 1));
+  } else if (phixs_file_version == 1) {
     globals::NPHIXSPOINTS = 100;
     globals::NPHIXSNUINCREMENT = .1;
-    last_phixs_nuovernuedge = 10;
+    last_phixs_nuovernuedge = 10;  // this matches classic
+    printout("using NPHIXSPOINTS = %d and NPHIXSNUINCREMENT = %lg set in input.cc to interpolate phixsdata.txt data\n",
+             globals::NPHIXSPOINTS, globals::NPHIXSNUINCREMENT);
```
lukeshingles, line 255: Accessing the file system every time a cross section is needed is going to cause serious performance issues. Replace the global int phixs_file_version with bool phixs_file_version_exists[] and just set them once in input.cc.

ratecoeff.cc
```diff
@@ -44,7 +45,8 @@ typedef struct {
 
 static char adatafile_hash[33];
 static char compositionfile_hash[33];
-static char phixsfile_hash[33];
+static char phixsfile_v1_hash[33];
+static char phixsfile_v2_hash[33];
```
lukeshingles, line 49: Put these in an array similar to phixsdata_filenames for easy looping. Also add an array for phixs_file_version_exists.

ratecoeff.cc
```diff
@@ -76,14 +78,42 @@ static bool read_ratecoeff_dat(void)
...
+    } else {
+      char phixsfile_v2_hash_in[33];
+      assert_always(fscanf(ratecoeff_file, "%32s\n", phixsfile_v2_hash_in) == 1);
+      printout("ratecoeff.dat: MD5 %s = %s ", phixsdata_filenames[2], phixsfile_v2_hash_in);
+      if (strcmp(phixsfile_v2_hash, phixsfile_v2_hash_in) == 0)
+        printout("(pass)\n");
+      else {
+        printout("\nMISMATCH: MD5 %s = %s\n", phixsdata_filenames[2], phixsfile_v2_hash);
+        fileisamatch = false;
+      }
```
lukeshingles, line 116: It looks like you've copied and pasted code. Don't repeat yourself and keep the code as simple as possible, for example:
```
for (int phixsver = 1; phixsver <= 2; phixsver++) {
    if (std::ifstream(phixsdata_filenames[phixsver]).good())) {
        char phixsfile_hash_in[33]
        assert_always(fscanf(ratecoeff_file, "%32s\n", phixsfile_hash_in) == 1);
        printout("ratecoeff.dat: MD5 %s = %s ", phixsdata_filenames[phixsver], phixsfile_hash_in);
        if (strcmp(phixsfile_hash[phixsver], phixsfile_hash_in) == 0) {
            printout("(pass)\n");
        } else {
           printout("\nMISMATCH: MD5 %s = %s\n", phixsdata_filenames[phixsver], phixsfile_hash[phixsver]);
           fileisamatch = false;
        }
    }
}
```

ratecoeff.cc
```diff
@@ -196,7 +226,16 @@ static void write_ratecoeff_dat(void) {
...
+  const bool phixs_v1_exists = std::ifstream(phixsdata_filenames[1]).good();
+  const bool phixs_v2_exists = std::ifstream(phixsdata_filenames[2]).good();
+  if (phixs_v1_exists && phixs_v2_exists) {
+    fprintf(ratecoeff_file, "%32s\n", phixsfile_v2_hash);
+    fprintf(ratecoeff_file, "%32s\n", phixsfile_v1_hash);
+  } else if (phixs_v1_exists) {
+    fprintf(ratecoeff_file, "%32s\n", phixsfile_v1_hash);
+  } else {
+    fprintf(ratecoeff_file, "%32s\n", phixsfile_v2_hash);
+  }
```
lukeshingles, line 238: Don't copy and paste code.
```
for (int phixsver = 1; phixsver <= 2; phixsver++) }
  if (phixs_file_version_exists[phixsver]) {
    fprintf(ratecoeff_file, "%32s\n", phixsfile_hash[phixsver])
  }
}
```

ratecoeff.cc
```diff
@@ -1006,7 +1045,16 @@ void ratecoefficients_init(void)
...
+  const bool phixs_v1_exists = std::ifstream(phixsdata_filenames[1]).good();
+  const bool phixs_v2_exists = std::ifstream(phixsdata_filenames[2]).good();
+  if (phixs_v1_exists && phixs_v2_exists) {
+    md5_file(phixsdata_filenames[1], phixsfile_v1_hash);
+    md5_file(phixsdata_filenames[2], phixsfile_v2_hash);
+  } else if (phixs_v1_exists) {
+    md5_file(phixsdata_filenames[1], phixsfile_v1_hash);
+  } else {
+    md5_file(phixsdata_filenames[2], phixsfile_v2_hash);
+  }
```
lukeshingles, line 1057:
```
for (int phixsver = 1; phixsver <= 2; phixsver++) }
  if (phixs_file_version_exists[phixsver]) {
    md5_file(phixsdata_filenames[phixsver], phixsfile_hash[phixsver])
  }
}
```

update_grid.cc
```diff
@@ -1016,8 +1016,10 @@ static void update_grid_cell(const int mgi, const int nts, const int nts_prev, c
 {
   const int assoc_cells = grid::get_numassociatedcells(mgi);
   if (assoc_cells > 0) {
-    // estimators were accumulated in nts_prev, but radiation density, etc should be scaled to the cell volume at
-    // nts
+    // estimators were accumulated in nts_prev so radiation density, etc should be scaled to the cell volume at
+    // nts_prev not nts to prevent the estimators being under estimated by scaling with a cell volume that is
+    // larger than the cell volume was when the estimators were being accumlated however still using
+    // nts for this release as don;t want to change the code behaviour with this release just adding functionality
```
lukeshingles, line 1022: Delete or rephrase into coherent sentences. The nts_prev change will be made in the next major release.

artisoptions_finn_nlte.h
lukeshingles, line 1: I think it would be better to describe what these options are useful for rather than naming it after yourself. This is early-time NLTE with no non-thermal solver? Call it something like artisoptions_nltewithoutnonthermal.h for example.

lukeshingles (changes requested):

input.cc
fionntancallan: I'm not sure I understand sorry, how does information on which phixsdata files exist from phixs_file_version_exists[] help determine which phixsdata file is being read on a given iteration of read_phixs_data() if both phixsdata files exist?

lukeshingles: Continue passing the phixs_file_version argument to read_phixs_data(). Just don't make it a global variable.

fionntancallan: phixs_file_version is used in the photoionization_crosssection_fromtable in atomic.cc. Should I retain a seperate global variable for phixs_file_version e.g. phixs_file_version_glob which I just use for this case?

lukeshingles: I suggested replacing that global variable with a new global bool phixs_file_version_exists[]. What number would phixs_file_version contain if both files are being used?

fionntancallan: if (phixs_file_version == 1) {
    // classic mode: no interpolation
    if (nu == nu_edge) {
      sigma_bf = photoion_xs[0];
    } else if (nu <= nu_edge * (1 + globals::NPHIXSNUINCREMENT * globals::NPHIXSPOINTS)) {
      int i = floor(nu / (globals::NPHIXSNUINCREMENT * nu_edge)) - 10;
      sigma_bf = photoion_xs[i];
    } else {
      /// use a parameterization of sigma_bf by the Kramers formula
      /// which anchor point should we take ??? the cross-section at the edge or at the highest grid point ???
      /// so far the highest grid point, otherwise the cross-section is not continuous
      sigma_bf = photoion_xs[globals::NPHIXSPOINTS - 1] *
                 pow(nu_edge * (1 + globals::NPHIXSNUINCREMENT * globals::NPHIXSPOINTS) / nu, 3);
    }
    return sigma_bf;
  }

Does this condition then not make sense if there are two phixsdata files being used? If what is done here is dependant on which phixsdata file is being used how would we then apply this condition?

ratecoeff.cc
fionntancallan: having trouble initialising the phixsfile_hash_array, I've tried a lot of different variations of this but none have worked:

__managed__ std::array<char 33, 3> phixsfile_hash;

I'm trying to make an array 3 elements long all with all elements having 32 characters

lukeshingles: You forgot the brackets around the number 33.

input.cc
fionntancallan: I declared phixs_file_version_exists in atomic.h like so:

extern __managed__ std::array<bool, 3> phixs_file_version_exists;

This is where phixs_file_version was previously declared and the atomic.h header is included in the files that use phixs_file_version_exists but I'm getting this error when compiling: 
 
undef: _phixs_file_version_exists
Undefined symbols for architecture x86_64:
  "_phixs_file_version_exists", referenced from:
      _main in lto.o
      photoionization_crosssection_fromtable(float const*, double, double) in lto.o
      read_atomicdata_files() in lto.o
      read_phixs_data(int) in lto.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [sn3d] Error 1

fionntancallan: I'm defining the boolean array elements like this incase that's relevant:

  phixs_file_version_exists[1] = std::ifstream(phixsdata_filenames[1]).good();
  phixs_file_version_exists[2] = std::ifstream(phixsdata_filenames[2]).good();

lukeshingles: Those are two assignments, not a definition. You need a definition that matches your declaration, so put 'std::array<bool, 3> phixs_file_version_exists;' in the cc file outside of any functions.
Ignore the managed keyword for now. It's for CUDA support but I'll be removing that in the next release.

fionntancallan: That worked thanks! Just testing that the code is behaving as expected for the different file configurations then I will update the pull request this afternoon.

fionntancallan:
Should I go ahead and squash and merge this pull request?

lukeshingles:
> Should I go ahead and squash and merge this pull request?

The last thing is that fstream header, then go ahead and merge.

PR #47, opened 2023-12-01
Add option to limit bf estimators

jpollin98:


lukeshingles:
A few things to do before we can merge:

- Set your IDE to use clang-format and pre-commit install to autoformat the code on commit.
- Rename BFGlobalVariable to something more descriptive like "bfestimcount". By moving the setup code from input.cc to radfield.cc, we can avoid introducing a new global variable altogether.
- Ensure that performance has not decreased in the default case of storing a full set of bf estimators. If performance is a problem, allcont can be split into two lists, with and without bf estimators to avoid the branching overhead.

jpollin98:
Hi Luke, I have made a few of the changes you have suggested. The code, for the most part, appears ready to merge into the main branch; once I have implemented the size_t changes, I agree some performance testing should be done. 

However, I have a question. For some reason, the code seemed to fall over on one of the Kilonova tests. However, when I reran the tests, it passed without causing any errors. Is this behaviour expected?

lukeshingles:
We need to add a new test for this functionality to ensure that it doesn't get broken in the future. I suggest just making a slightly modified version of the nebularonezone test. Do you want me to do that part or do you want to try it?

lukeshingles:
> Hi Luke, I have made a few of the changes you have suggested. The code, for the most part, appears ready to merge into the main branch; once I have implemented the size_t changes, I agree some performance testing should be done.
> 
> However, I have a question. For some reason, the code seemed to fall over on one of the Kilonova tests. However, when I reran the tests, it passed without causing any errors. Is this behaviour expected?

Yes, I have seen this before on the kilonova test. I'm not sure exactly how it happens, but I suspect that we occasionally get assigned a runner with a slightly different CPU or software environment and this changes the checksums. Re-running makes sense in that situation.

jpollin98:
> We need to add a new test for this functionality to ensure that it doesn't get broken in the future. I suggest just making a slightly modified version of the nebularonezone test. Do you want me to do that part or do you want to try it?

I'm happy to try to make a new nebularonezone test that tests the functionality, as it's valuable to me to be more competent at code development.

lukeshingles:
> > We need to add a new test for this functionality to ensure that it doesn't get broken in the future. I suggest just making a slightly modified version of the nebularonezone test. Do you want me to do that part or do you want to try it?
> 
> I'm happy to try to make a new nebularonezone test that tests the functionality, as it's probably valuable for me to be a bit more competent at the code development side of things.

Ok, great. It's currently only in the develop branch, but take a look at how kilonova_2d_2dgrid_expansionopac is derived from kilonova_2d_2dgrid for an example.

lukeshingles (approved):

PR #56, opened 2024-05-07
Barnes thermalization

gleck97:


gammapkt.h
```diff
@@ -7,7 +7,7 @@
 namespace gammapkt {
 void init_gamma_data();
 void pellet_gamma_decay(Packet &pkt);
-void do_gamma(Packet &pkt, double t2);
+void treat_gamma_packet(Packet &pkt, double t2);
```
lukeshingles, line 10: What is the reason for changing this? This should be consistent with do_rpkt, do_ntlepton, do_kpkt, etc.

gleck97: I've just renamed the functions to make it consistent again

PR #62, opened 2024-05-17
Extend electron binding energies and shell occupancy data for work function approximation of non-thermal ionisation

ccollins22:


nonthermal.cc
```diff
@@ -155,30 +158,114 @@ static struct nt_solution_struct *nt_solution;
...
+}
+
+inline bool exists_test (const std::string& name) {
+  if (FILE *file = fopen(name.c_str(), "r")) {
+    fclose(file);
+    return true;
+  } else {
+    return false;
+  }
+}
```
lukeshingles, line 202: Is there a reason not to use std::filesystem::exists() here?

ccollins22: Just didn't know about it. Feel free to replace with whatever you like

lukeshingles (approved):

lukeshingles (approved):

lukeshingles (approved):

PR #71, opened 2024-05-27
Just opacity in grid.cc

gleck97:
Added the Just opacity used in ALCAR simulations that I implemented some time ago already


lukeshingles:
Looks great. I'm just about to commit a bug fix (should only affect Type Ia classic models hopefully) and then we can merge this one.

PR #76, opened 2024-05-31
Add Barnes Particle Thermalisation

gleck97:
Add Barnes thermalisation for alpha and beta particles from today's meeting


update_packets.cc
```diff
@@ -29,10 +29,39 @@ namespace {
...
+             PARTICLE_THERMALISATION_SCHEME == ThermalisationScheme::BARNES_LOCAL) {
+    const double t_0 = grid::get_t_model();
+    double rho_0 = 0.;
+    const double endot_per_rho = (pkt.pellet_decaytype == decay::DECAYTYPE_ALPHA) ? 5.e11 * MEV : 4.e10 * MEV;
+    if (PARTICLE_THERMALISATION_SCHEME == ThermalisationScheme::BARNES_GLOBAL) {
+      double V_0 = 0.;
+      for (int nonemptymgi = 0; nonemptymgi < grid::get_nonempty_npts_model(); nonemptymgi++) {
+        const int mgi = grid::get_mgi_of_nonemptymgi(nonemptymgi);
+        V_0 += grid::get_modelcell_assocvolume_tmin(mgi);
+      }
```
lukeshingles, line 46: I see two possible mistakes here.
- This volume does not include the volume of empty cells. Perhaps you want something like 4/3 * pi * (globals::vmax *globals::tmin)^3 for a spherical approximation of volume? or (2 * vmax * globals::tmin)^3 for the full cube.
- V_0 is the volume of non-empty cells at tmin (the simulation start time). Then rho_0 is the density at tmin. But in the expression for C, why is rho_0 multiplied by t_model^3? I can't follow the logic.

artisoptions_kilonova_lte.h
```diff
@@ -146,6 +144,8 @@ constexpr float RPKT_BOUNDBOUND_THERMALISATION_PROBABILITY = -1.;
 
 constexpr bool USE_XCOM_GAMMAPHOTOION = false;
 
+constexpr auto PARTICLE_THERMALISATION_SCHEME = ThermalisationScheme::INSTANT;
```
lukeshingles, line 147: To preserve the exiting behaviour, the thermalisation scheme should be ThermalisationScheme::DETAILED. All other option files need to be updated to use the existing instant mode. (small detail, maybe we can switch to Title case like Instant, Detailed, BarnesGlobal, BarnesLocal).

lukeshingles (changes requested):

update_packets.cc
lukeshingles: This also looks totally unfamiliar to me from when I calculated the Barnes thermalisation efficiencies for plotting in artistools. Why are they so different?

```python
        ejecta_ke = (0.5 * (dfmodel["mass_g"] / 1000.0) * (dfmodel["vel_r_mid"] / 100.0) ** 2).sum()
        # velocity derived from ejecta kinetic energy to match Barnes et al. (2016) Section 2.1
        ejecta_v = np.sqrt(2 * ejecta_ke / (model_mass_grams * 1e-3))
        v2 = ejecta_v / (0.2 * 299792458)
        print(f"  Barnes average ejecta velocity: {ejecta_v / 299792458:.2f}c")
        m5 = model_mass_grams / (5e-3 * 1.989e33)  # M / (5e-3 Msun)

        e0_beta_mev = 0.5
        # Barnes et al (2016) equation 20
        t_ineff_beta = 7.4 * (e0_beta_mev / 0.5) ** -0.5 * m5**0.5 * (v2 ** (-3.0 / 2))
        # Barnes et al (2016) equation 32
        barnes_f_beta = math.log(1 + 2 * (t / t_ineff_beta) ** 2) / (2 * (t / t_ineff_beta) ** 2)
        
        e0_alpha_mev = 6.0
        # Barnes et al (2016) equation 25 times equation 16 for t_peak
        t_ineff_alpha = 4.3 * 1.8 * (e0_alpha_mev / 6.0) ** -0.5 * m5**0.5 * (v2 ** (-3.0 / 2))
        # Barnes et al (2016) equation 32
        barnes_f_alpha =math.log(1 + 2 * (t / t_ineff_alpha) ** 2) / (2 * (t / t_ineff_alpha) ** 2)
```

artisoptions_kilonova_lte.h
gleck97: Yes, thanks. Will push that along with changes to your comment above.

update_packets.cc
gleck97: You're right in both points. In the first case, I think I have to choose the spherical approximation (which is used by Barnes), in the second point that's just a mistake, I need a reference time / density where which has to be consistent.

gleck97: > I see two possible mistakes here.
> 
>     * This volume does not include the volume of empty cells. Perhaps you want something like 4/3 * pi * (globals::vmax *globals::tmin)^3 for a spherical approximation of volume? or (2 * vmax * globals::tmin)^3 for the full cube.
> 
>     * V_0 is the volume of non-empty cells at tmin (the simulation start time). Then rho_0 is the density at tmin. But in the expression for C, why is rho_0 multiplied by t_model^3? I can't follow the logic.

You're right in both points. In the first case, I think I have to choose the spherical approximation (which is used by Barnes), in the second point that's just a mistake, I need a reference time / density where which has to be consistent.

gleck97: > This also looks totally unfamiliar to me from when I calculated the Barnes thermalisation efficiencies for plotting in artistools. Why are they so different?
> 
> ```python
>         ejecta_ke = (0.5 * (dfmodel["mass_g"] / 1000.0) * (dfmodel["vel_r_mid"] / 100.0) ** 2).sum()
>         # velocity derived from ejecta kinetic energy to match Barnes et al. (2016) Section 2.1
>         ejecta_v = np.sqrt(2 * ejecta_ke / (model_mass_grams * 1e-3))
>         v2 = ejecta_v / (0.2 * 299792458)
>         print(f"  Barnes average ejecta velocity: {ejecta_v / 299792458:.2f}c")
>         m5 = model_mass_grams / (5e-3 * 1.989e33)  # M / (5e-3 Msun)
> 
>         e0_beta_mev = 0.5
>         # Barnes et al (2016) equation 20
>         t_ineff_beta = 7.4 * (e0_beta_mev / 0.5) ** -0.5 * m5**0.5 * (v2 ** (-3.0 / 2))
>         # Barnes et al (2016) equation 32
>         barnes_f_beta = math.log(1 + 2 * (t / t_ineff_beta) ** 2) / (2 * (t / t_ineff_beta) ** 2)
>         
>         e0_alpha_mev = 6.0
>         # Barnes et al (2016) equation 25 times equation 16 for t_peak
>         t_ineff_alpha = 4.3 * 1.8 * (e0_alpha_mev / 6.0) ** -0.5 * m5**0.5 * (v2 ** (-3.0 / 2))
>         # Barnes et al (2016) equation 32
>         barnes_f_alpha =math.log(1 + 2 * (t / t_ineff_alpha) ** 2) / (2 * (t / t_ineff_alpha) ** 2)
> ```

Those equations are for specific values of M_ej etc., right? I thought it would be better to use the general formula which leaves M_ej, v_ej as parameters. But I still tried to use the basic assumption of Barnes of a uniform sphere

gleck97:
Could you have a look at the changes I pushed concerning the "Wollaeger" scheme? In gammapkt.cc, starting from line 1038, I added basically the same commands as from vpkt.cc (line 260) to obtain the optical depth of a packet outwards in radial direction via an integration / summation over cells. I tried so via a copy of the current packet, but the whole thing causes a segmentation fault. Do you have an idea why?

lukeshingles:
> Could you have a look at the changes I pushed concerning the "Wollaeger" scheme? In gammapkt.cc, starting from line 1038, I added basically the same commands as from vpkt.cc (line 260) to obtain the optical depth of a packet outwards in radial direction via an integration / summation over cells. I tried so via a copy of the current packet, but the whole thing causes a segmentation fault. Do you have an idea why?

Please make a separate PR for the Wollaeger scheme (we should also discuss it in a meeting so I can understand how this is supposed to work). I want to get the Barnes thermalisation work finished and merged since it touches some of the same functions as my PR #86.

lukeshingles (approved):

PR #82, opened 2024-06-03
Reduce bfrate raw on each node and Broadcast reduced quantity; Replace nt_bcast with MPI Pack/Unpack

jpollin98:
I have replaced the nt_bcast of the frac_deposition, ratecoeffperdeposition, and lineindex with MPI_Pack and MPI_Unpack due to the significant number of broadcasts in a 3D nebular model for each grid cell producing MPI errors.

The bfrate_raw MPI_Allreduce, when performed as a loop over all grid cells on all nodes, is a significant bottleneck for 3D Nebular models. The change implemented is as follows: 
For each nonempty cell, it reduces values across all processes within the same node. We then reduce across all nodes, and finally, the results are broadcast back to all processes within each node.


lukeshingles:
Please target your PRs to the develop branch (in case we need to do large scale tests to check correctness before main branch gets synchronised). Now that #78 is merged, you can rebase your branch and should get a fairly reliable test for multi-node MPI behaviour.

nonthermal.cc
```diff
@@ -2692,15 +2692,49 @@ void nt_MPI_Bcast(const int modelgridindex, const int root) {
     }
 
     const auto frac_excitations_list_size = nt_solution[modelgridindex].frac_excitations_list.size();
+    int buffer_size = (2 * sizeof(double) + sizeof(int)) * frac_excitations_list_size;
+    char *buffer = (char *)malloc(buffer_size);
+    int position = 0;
+
+    printout("Size of frac_excitations_list: %zu\n", frac_excitations_list_size);
+    printout("Buffer size allocated for fractional excitations is %g Mb\n", buffer_size / 1024. / 1024.);
```
lukeshingles, line 2700: That should be capital 'B' for byte, not 'b' for bit, right?

nonthermal.cc
```diff
@@ -2692,15 +2692,49 @@ void nt_MPI_Bcast(const int modelgridindex, const int root) {
     }
 
     const auto frac_excitations_list_size = nt_solution[modelgridindex].frac_excitations_list.size();
+    int buffer_size = (2 * sizeof(double) + sizeof(int)) * frac_excitations_list_size;
+    char *buffer = (char *)malloc(buffer_size);
```
lukeshingles, line 2696: Please replace the C-style cast with C++ static_cast<char *>.

lukeshingles:
Passed the tests on the first try. Nice work! I would just request two very minor changes and then it's good to merge.

lukeshingles (changes requested):
Passed on the first try. Nice work! I would just request two very minor changes and then it's good to merge into develop.

radfield.cc
```diff
@@ -1149,9 +1149,15 @@ void reduce_estimators()
...
-                    MPI_SUM, MPI_COMM_WORLD);
+      MPI_Reduce(globals::rank_in_node == 0 ? MPI_IN_PLACE : &bfrate_raw[nonemptymgi * globals::bfestimcount],
+                 &bfrate_raw[nonemptymgi * globals::bfestimcount], globals::bfestimcount, MPI_DOUBLE, MPI_SUM, 0,
+                 globals::mpi_comm_node);
     }
+    if (globals::rank_in_node == 0) {
+      MPI_Allreduce(MPI_IN_PLACE, bfrate_raw.data(), nonempty_npts_model * globals::bfestimcount, MPI_DOUBLE, MPI_SUM,
+                    globals::mpi_comm_internode);
+    }
+    MPI_Bcast(bfrate_raw.data(), nonempty_npts_model * globals::bfestimcount, MPI_DOUBLE, 0, globals::mpi_comm_node);
```
lukeshingles, line 1160: Sorry, this just occured to me. An MPI_Allreduce followed by an MPI_Bcast should be less efficient than an MPI_Allreduce. Would it be better to replace the loop of MPI_Allreduce calls over each cell with just one since the entire array needs to be reduced? This code is there because it used to selectively reduce only non-empty cells and skip over empty ones, but that's not needed now that only non-empty cells get stored in the array.

lukeshingles: Actually, there might be a reason to keep that loop. Sending a size of (nonempty_npts_model * globals::bfestimcount) could overflow the 32-bit int size parameter of the MPI call.

lukeshingles (approved):

PR #92, opened 2024-06-17
Fix Wollaeger Thermalisation typo

gleck97:
Removed the same (copying) error as in Bulla 2023


lukeshingles:
Please make a comment with a reference to the Bulla paper so that we can keep track of this.

lukeshingles (approved):

PR #96, opened 2024-07-25
Add switch for calling Te solver

ccollins22:
When heating and cooling rates are too noisy to calculate reliable Te (e.g. only a few species are NLTE) add option to set Te=TJ instead


lukeshingles:
I didn't understand the physics justification, but on the software side, is it worth the maintenance burden of an extra configuration option versus four lines of local changes? Are you going to publish models with this mode?

PR #99, opened 2024-08-05
Use node-shared memory for non-thermal excitation ratecoeffs

jpollin98:
The main NT_EXCITATIONS list is now stored on node memory instead of local memory as dynamically allocating memory on each rank requires an extremely large amount of memory for 3D nebular simulations.

### DIFF
--- a/nonthermal.cc
+++ b/nonthermal.cc
@@ -1635,19 +1635,19 @@ void analyse_sf_solution(const int modelgridindex, const int timestep, const boo
     const int ntransdisplayed = std::min(50, nt_solution[modelgridindex].frac_excitations_list_size);
 
     for (excitationindex = 0; excitationindex < ntransdisplayed; excitationindex++) {
-      const double frac_deposition = nt_solution[modelgridindex].frac_excitations_list[excitationindex].frac_deposition;
+      const auto &ntexc = nt_solution[modelgridindex].frac_excitations_list[excitationindex];
+      const double frac_deposition = ntexc.frac_deposition;
       if (frac_deposition > 0.) {
-        const int lineindex = nt_solution[modelgridindex].frac_excitations_list[excitationindex].lineindex;
-        const TransitionLine *line = &globals::linelist[lineindex];
-        const int element = line->elementindex;
-        const int ion = line->ionindex;
-        const int lower = line->lowerlevelindex;
-        const int upper = line->upperlevelindex;
-        const int uptransindex = nt_solution[modelgridindex].frac_excitations_list[excitationindex].loweruptransindex;
+        const int lineindex = ntexc.lineindex;
+        const TransitionLine &line = globals::linelist[lineindex];
+        const int element = line.elementindex;
+        const int ion = line.ionindex;
+        const int lower = line.lowerlevelindex;
+        const int upper = line.upperlevelindex;
+        const int uptransindex = ntexc.loweruptransindex;
         const double epsilon_trans = epsilon(element, ion, upper) - epsilon(element, ion, lower);
 
-        const double ratecoeffperdeposition =
-            nt_solution[modelgridindex].frac_excitations_list[excitationindex].ratecoeffperdeposition;
+        const double ratecoeffperdeposition = ntexc.ratecoeffperdeposition;
         const double ntcollexc_ratecoeff = ratecoeffperdeposition * deposition_rate_density;
 
         const double t_mid = globals::timesteps[timestep].mid;
--- a/nonthermal.cc
+++ b/nonthermal.cc
@@ -137,6 +137,9 @@ thread_local std::vector<nt_excitation_struct> tmp_excitation_list;
 // pointer to either local or node-shared memory excitation list of all cells
 nt_excitation_struct *excitations_list_all_cells{};
 
+// the minimum of MAX_NT_EXCITATIONS_STORED and the number of included excitation transitions in the atomic dataset
+int nt_excitations_stored = 0;
+
 struct nt_solution_struct {
   double *yfunc{};  // Samples of the Spencer-Fano solution function. Multiply by energy to get non-thermal
                     // electron number flux. y(E) * dE is the flux of electrons with energy in the range (E, E +
@@ -462,6 +465,27 @@ void read_collion_data() {
   if (NT_MAX_AUGER_ELECTRONS > 0) {
     read_auger_data();
   }
+}
+
+auto get_possible_nt_excitation_count() -> int {
+  int ntexcitationcount = 0;
+  for (int element = 0; element < get_nelements(); element++) {
+    for (int ion = 0; ion < get_nions(element); ion++) {
+      const int nlevels = std::min(NTEXCITATION_MAXNLEVELS_LOWER, get_nlevels(element, ion));
+      for (int lower = 0; lower < nlevels; lower++) {
+        const int nuptrans = get_nuptrans(element, ion, lower);
+        for (int t = 0; t < nuptrans; t++) {
+          const int lineindex = globals::elements[element].ions[ion].levels[lower].uptrans[t].lineindex;
+          const int upper = globals::linelist[lineindex].upperlevelindex;
+          if (upper >= NTEXCITATION_MAXNLEVELS_UPPER) {
+            continue;
+          }
+          ntexcitationcount++;
+        }
+      }
+    }
+  }
+  return ntexcitationcount;
 }
 
 void zero_all_effionpot(const int modelgridindex) {
@@ -1579,18 +1603,18 @@ void analyse_sf_solution(const int modelgridindex, const int timestep, const boo
     }
   }
 
-  if constexpr (NT_EXCITATION_ON && (MAX_NT_EXCITATIONS_STORED > 0)) {
+  if (nt_excitations_stored > 0) {
     // sort by descending frac_deposition
     std::ranges::stable_sort(tmp_excitation_list, std::ranges::greater{}, &nt_excitation_struct::frac_deposition);
 
     // the excitation list is now sorted by frac_deposition descending
     const double deposition_rate_density = get_deposition_rate_density(modelgridindex);
 
-    if (std::ssize(tmp_excitation_list) > MAX_NT_EXCITATIONS_STORED) {
+    if (std::ssize(tmp_excitation_list) > nt_excitations_stored) {
       // truncate the sorted list to save memory
       printout("  Truncating non-thermal excitation list from %zu to %d transitions.\n", tmp_excitation_list.size(),
-               MAX_NT_EXCITATIONS_STORED);
-      tmp_excitation_list.resize(MAX_NT_EXCITATIONS_STORED);
+               nt_excitations_stored);
+      tmp_excitation_list.resize(nt_excitations_stored);
     }
 
     std::copy(tmp_excitation_list.begin(), tmp_excitation_list.end(),
@@ -1963,11 +1987,13 @@ void init(const int my_rank, const int ndo_nonempty) {
   }
 
   if (NT_EXCITATION_ON) {
-    printout(
-        "[info] mem_usage: storing non-thermal frac_excitations_list for all allocated cells occupies %.3f MB\n",
-        grid::get_nonempty_npts_model() * sizeof(nt_excitation_struct) * MAX_NT_EXCITATIONS_STORED / 1024. / 1024.);
+    nt_excitations_stored = std::min(MAX_NT_EXCITATIONS_STORED, get_possible_nt_excitation_count());
+    printout("[info] mem_usage: storing %d non-thermal excitations for non-empty cells occupies %.3f MB\n",
+             nt_excitations_stored,
+             grid::get_nonempty_npts_model() * sizeof(nt_excitation_struct) * nt_excitations_stored / 1024. / 1024.);
 
     const ptrdiff_t nonempty_npts_model = grid::get_nonempty_npts_model();
+
 #ifdef MPI_ON
 
     MPI_Win win_shared_excitations_list{};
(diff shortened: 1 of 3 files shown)


nltepop.cc
lukeshingles, line 1: These changes seem to have been included by mistake. Please remove them from the PR.

nonthermal.cc
```diff
@@ -467,7 +467,64 @@ void init(const int my_rank, const int ndo_nonempty) {
     fflush(nonthermalfile);
   }
 
+  int LIST_LENGTH = MAX_NT_EXCITATIONS_STORED;
```
lukeshingles, line 470: What is the purpose of this alias?

nonthermal.cc
```diff
@@ -467,7 +467,64 @@ void init(const int my_rank, const int ndo_nonempty) {
...
+  // We want to keep the nt_solution_struct in local but the nt_excitation_struct in shared memory
+  int rank, shm_rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+  // Get the rank of the shared memory communicator
+  MPI_Comm shmcomm;
+  MPI_Comm_split_type(MPI_COMM_WORLD, MPI_COMM_TYPE_SHARED, 0, MPI_INFO_NULL, &shmcomm);
+
+  // Rank within the shared memory communicator
+  MPI_Comm_rank(shmcomm, &shm_rank);
```
lukeshingles, line 484: These lines are redundant since we already have the rank in globals::rank_global and the shared memory communicator is globals::mpi_comm_node. It's especially important to use globals::mpi_comm_node since on the CI jobs, these "nodes" are further split into subsets of the shared memory system in order to test internode and intranode communication.

nonthermal.cc
```diff
@@ -467,7 +467,64 @@ void init(const int my_rank, const int ndo_nonempty) {
...
+  int list_size = grid::get_npts_model() * sizeof(nt_excitation_struct) * LIST_LENGTH;
+
+  // Now add a printout telling us how much memory we are using / going to use
+  printout("[info] mem_usage: storing non-thermal frac_excitations_list for all allocated cells occupies %.3f MB\n",
+           list_size / 1024. / 1024.);
+
+  // MPI_Aint is the type of a variable able to contain a memory address.
+  MPI_Aint shared_mem_size = (shm_rank == 0) ? list_size : 0;
+
+  // Allocate shared memory for frac_excitations_list on local rank 0
```
lukeshingles, line 496: This might be ok (and simplifies the code), but I always worried that some clusters would failed to account for the memory properly when everything is assigned to node_rank 0. In other parts of the code, we distribute the array as evenly as possible.

nonthermal.cc
```diff
@@ -467,7 +467,64 @@ void init(const int my_rank, const int ndo_nonempty) {
...
+  nt_solution = static_cast<nt_solution_struct *>(calloc(grid::get_npts_model(), sizeof(nt_solution_struct)));
+
+  // Now we need to set the pointers for the nt_excitation_struct inside the nt_solution_struct to the shared memory
+  // Calculate how many excitations per nt_solution_struct
+  int excitations_per_solution = LIST_LENGTH;
+
+  for (int modelgridindex = 0; modelgridindex < grid::get_npts_model(); modelgridindex++) {
+    nt_excitation_struct *start = &shared_excitations_list[modelgridindex * excitations_per_solution];
+    nt_excitation_struct *end = &shared_excitations_list[((modelgridindex + 1) * excitations_per_solution) - 1];
+    nt_solution[modelgridindex].frac_excitations_list.assign(start, end);
```
lukeshingles, line 517: This is making a copy, so it looks like we are occupying the same memory per node as before (have you checked this in the JUWELS job report? did memory usage decrease?) plus the shared memory copy. Change the type of frac_excitations_list from std::vector to std::span so that it can hold pointers to the shared array instead of having its own storage with a copy of it.

nonthermal.cc
```diff
@@ -2653,7 +2710,8 @@ void read_restart_data(FILE *gridsave_file) {
 }
 
 #ifdef MPI_ON
-void nt_MPI_Bcast(const int modelgridindex, const int root) {
+void nt_MPI_Bcast(const int modelgridindex, const int root, const int my_rank, char *nt_maxstored_buffer,
+                  int buffer_size) {
```
lukeshingles, line 2714: const int buffer_size

nonthermal.cc
```diff
@@ -2692,15 +2750,46 @@ void nt_MPI_Bcast(const int modelgridindex, const int root) {
...
+      printout("Broadcasting for MGI = %d\n", modelgridindex);
+    }
+
+    MPI_Bcast(nt_maxstored_buffer, buffer_size, MPI_PACKED, root, MPI_COMM_WORLD);
+
+    MPI_Barrier(MPI_COMM_WORLD);
+
+    position = 0;
+
     for (size_t excitationindex = 0; excitationindex < frac_excitations_list_size; excitationindex++) {
```
lukeshingles, line 2779: It would be nice if we could eliminate the complexity of the MPI_Packing and extra buffer after the nt_excitations list has been de-duplicated on the nodes. Check whether this is really necessary.

lukeshingles (changes requested):

nonthermal.cc
jpollin98: Was part of the printouts I had for checking allocations etc. I will remove / clean up the code

nltepop.cc
jpollin98: removed changes to nltepop.cc

nonthermal.cc
jpollin98: Changed allocation of memory from using shmcomm to globals::rank_in_node

nonthermal.cc
jpollin98: So the method of allocation onto rank 0 can cause problems because it might unevenly allocate memory on the node ? Subsequently what is the method for distributing the array evenly (If you can point to a similar example I am happy to try and implement that method)

nonthermal.cc
jpollin98: I’m most confused about this part: As I understand it, the current issue with the code arises during the frac_excitations_list resizing on the clusters during the nt_MPI_Bcast.  Each rank could theoretically require memory for MAX_NT_EXCITATIONS_STORED * the Total number of grid cells * the size of the vector struct. Hence, we are running out of memory.

Stuarts suggested fix: place the vector into shared memory, so that only one copy of each is needed per node, rather than per rank. Although this will require a very large structure to be allocated on the node, it should solve the memory issue. Hence after moving the structure into shared memory, we will need to assign pointers so that each rank knows the location of each cell in this large shared list. I don’t think the current implementation is doing this correctly then…

Another question on this section is does std::span when clear is applied does this remove the pointers to shared memory. I think that is the case when vectors are cleared.

nonthermal.cc
jpollin98: Resolved const int buffer_size problem

nonthermal.cc
jpollin98: Theoretically if the above is implemented, we could just reduce the quantity across each rank to 0 instead of casting.

nonthermal.cc
lukeshingles: I never confirmed it, but we often we request a certain amount of memory "per task" and I'm not sure if one of the tasks exceeding that limit will cause a problem (or whether the memory of all tasks on a node is correctly averaged out).

If you want more examples, just search the code for MPI_Win_allocate_shared.

nonthermal.cc
```diff
@@ -2683,22 +2739,16 @@ void nt_MPI_Bcast(const int modelgridindex, const int root) {
...
-      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list[excitationindex].frac_deposition, 1, MPI_DOUBLE,
-                root, MPI_COMM_WORLD);
-      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list[excitationindex].ratecoeffperdeposition, 1,
-                MPI_DOUBLE, root, MPI_COMM_WORLD);
-      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list[excitationindex].lineindex, 1, MPI_INT, root,
-                MPI_COMM_WORLD);
+    if (globals::rank_in_node == 0) {
+      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list,
+                nt_solution[modelgridindex].frac_excitations_list_size, MPI_FLOAT, root_node_id,
+                globals::mpi_comm_internode);
```
lukeshingles, line 2751: frac_excitations_list is a pointer to a nt_excitation_struct, not a pointer to a float. To do the Bcast, you should be able to set the data type to MPI_BYTE and then set the length to frac_excitations_list_size * sizeof(nt_excitation_struct).

nonthermal.cc
```diff
@@ -2727,4 +2777,4 @@ void nt_print_stats(const double modelvolume, const double deltat) {
   printout("nt_energy_deposited = %g [eV/s/cm^3]\n", deposition_rate_density_montecarlo);
 }
 
-}  // namespace nonthermal
\ No newline at end of file
+}  // namespace nonthermal
```
lukeshingles, line 2780: Check that you have pre-commit installed correctly, which should make sure that all text files end with a single newline.

sn3d.cc
```diff
@@ -597,8 +597,7 @@ void zero_estimators() {
   MPI_Barrier(MPI_COMM_WORLD);
 #endif
 }
-
-auto do_timestep(const int nts, const int titer, const int my_rank, const int nstart, const int ndo, Packet *packets,
+auto do_timestep(int nts, const int titer, const int my_rank, const int nstart, const int ndo, Packet *packets,
```
lukeshingles, line 600: Mistake?

sn3d.cc
```diff
@@ -629,7 +628,9 @@ auto do_timestep(const int nts, const int titer, const int my_rank, const int ns
 
   const auto sys_time_start_communicate_grid = std::time(nullptr);
 
-/// Each process has now updated its own set of cells. The results now need to be communicated between processes.
+  /// Each process has now updated its own set of cells. The results now need to be communicated between processes.
+  // allocate a buffer to broadcast MPI quantities
+
```
lukeshingles, line 633: Included by mistake?

radfield.cc
```diff
@@ -1140,9 +1140,15 @@ void reduce_estimators()
...
-                    MPI_SUM, MPI_COMM_WORLD);
+      MPI_Reduce(globals::rank_in_node == 0 ? MPI_IN_PLACE : &bfrate_raw[nonemptymgi * globals::bfestimcount],
+                 &bfrate_raw[nonemptymgi * globals::bfestimcount], globals::bfestimcount, MPI_DOUBLE, MPI_SUM, 0,
+                 globals::mpi_comm_node);
     }
+    if (globals::rank_in_node == 0) {
+      MPI_Allreduce(MPI_IN_PLACE, bfrate_raw.data(), nonempty_npts_model * globals::bfestimcount, MPI_DOUBLE, MPI_SUM,
+                    globals::mpi_comm_internode);
+    }
+    MPI_Bcast(bfrate_raw.data(), nonempty_npts_model * globals::bfestimcount, MPI_DOUBLE, 0, globals::mpi_comm_node);
```
lukeshingles, line 1151: Were this changes included by mistake? (they don't seem related to NT excitations). The previous version here should be much more efficient (and less code too!).

nltepop.cc
```diff
@@ -1228,4 +1228,4 @@ void nltepop_read_restart_data(FILE *restart_file) {
         assert_always(fscanf(restart_file, "%la ", &grid::modelgrid[modelgridindex].nlte_pops[nlteindex]) == 1);
     }
   }
-}
\ No newline at end of file
+}
```
lukeshingles, line 1231: Run pre-commit install.

nonthermal.cc
```diff
@@ -124,7 +125,6 @@ struct nt_excitation_struct {
   int lineindex;
   int loweruptransindex;
 };
-
```
lukeshingles, line 127: Why did this line break disappear?

radfield.cc
jpollin98: Yeah, sorry, I had started editing this branch with other changes in it (a rookie mistake). Thats why there were so many strange changes!

nonthermal.cc
```diff
@@ -2682,23 +2739,17 @@ void nt_MPI_Bcast(const int modelgridindex, const int root) {
...
-      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list[excitationindex].ratecoeffperdeposition, 1,
-                MPI_DOUBLE, root, MPI_COMM_WORLD);
-      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list[excitationindex].lineindex, 1, MPI_INT, root,
-                MPI_COMM_WORLD);
+    if (globals::rank_in_node == 0) {
+      // communicate NT excitation list via inter-node communication
+      MPI_Bcast(&nt_solution[modelgridindex].frac_excitations_list,
+                nt_solution[modelgridindex].frac_excitations_list_size * sizeof(nt_excitation_struct), MPI_BYTE,
+                root_node_id, globals::mpi_comm_internode);
     }
```
jpollin98, line 2753: There is something wrong with this bcast. I have run a 1D model, and on timestep 8, on the first time going through this function, the code crashes. Is there something simple thats wrong with this bcast that I'm missing? There are sensible values inside the frac_excitations_list, so its not that the values are infs etc

lukeshingles: It looks like there’s an extra ampersand on the first argument.

lukeshingles: I’m surprised that this didn’t crash the CI job…it’s worth figuring that out because we probably need a better test.

jpollin98: That fixed the problem. I had assumed we needed the ampersand to point to a specific memory address, but apparently not.

lukeshingles: frac_excitations_list is a already pointer type, so taking the address of it gives you a pointer to the pointer.

lukeshingles (approved):

lukeshingles:
I've backed up your branch to mpi_pack_log_test_backup, and re-applied your changes onto the latest develop commit since there were some conflicts.

PR #101, opened 2024-08-07
Include neglect expansion option in grid boundary dist calc

gleck97:


lukeshingles:
This is an approximate way to neglect expansion (using a faster speed of light) and affects a critical part of the code. if you can use templates to avoid any performance cost and really remove the expansion factors, then I could merge.

PR #105, opened 2024-08-15
Communicate elemental abundance updates between ranks

lukeshingles:
Elemental abundances are set to the snapshot values for all ranks on startup. The elemental abundance changes due to radioactive decays for each cell are calculated by one of the ranks, and then not propagated to all ranks before packet propagation. The elemental mass fractions were not used in many places (the ion and level populations were based on the correct updated elemental abundances and communicated), but do get checked for being either zero or greater than zero for certain features, which could be important with elemental abundances of zero the input snapshot that become non-zero due to radioactive decays.

Consequences:
- A zero abundance check will disable detailed bound-free estimator accumulation on all but one rank if the initial abundance is 0.0 while the true abundance is >0.
- For small (<MINPOP) populations in calculate_levelpop() and get_groundlevelpop(), a non-zero elemental abundance returns MINPOP, and a zero elemental abundance returns 0.
- The work function approximation for non-thermal ionisation involved an average atomic number (Zbar) that is not up to date. This probably has a minor effect on macroatom statistics, since Zbar will not change too much, and the ionisation balance was set by the rank doing the grid update with correct abundances.
- The unpublished opacity case 6 (to match Just et al. 2022) depends on a calculation of the lanthanide mass fraction that will be incorrect (derived froms initial composition instead of current one) for cell not updated by the current rank.
- The unpublished USE_XCOM_GAMMAPHOTOION mode will use initial elemental abundances instead of updated values to calculate photoelectric effect cross sections.

From the checksums of the tests, it appears that only nebular runs with detailed bound-free estimators are affected (very slightly, probably from the minor work-function change for Z=27 ionstage 4 that has no NT ionisation cross section available). A full-scale nebular test with a Shen+2018 model is in progress.


lukeshingles:
Spectral difference was negligible for Shen+2018 at 300 days.
plotspec_300.2d_318.9d.pdf

PR #120, opened 2024-09-02
Refix update_packets.cc and improve notation

gleck97:
Checksums have to be taken back as well


lukeshingles:
Does this version agree with the Barnes thermalisation calculations in artistools?

PR #131, opened 2024-10-01
Fix directory error

gleck97:


lukeshingles:
The previous code was correct and sn3d should not read files in the repository. If you need to use this file, then copy it into the model folder, or modelfolder/data.

PR #145, opened 2024-11-05
Get constant neutrino loss fraction

gleck97:


lukeshingles:
You will need to adjust the energy fractions just after the gamma-ray tables have been read. You can rescale the gamma and beta-particle energy per decay based on the Q values and your target neutrino fraction.

lukeshingles:
I don't see a reason for bool packet_setup_phase. I think this will lead to inconsistencies.

lukeshingles:
If you guard this functionality behind a new feature flag, you can use the CI system to detect some kinds of errors.

PR #159, opened 2024-11-22
Add more robust checks on NLTE population solver and safer treatment when NLTE solver fails

fionntancallan:
Add more robust checks on the populations in the NLTE solver and safer treatment of what populations should be used instead if the NLTE solution is not sensible (e.g. convert to LTE). In particular:

 -Previously when the NLTE population was negative for a level it was directly replaced with the LTE level population. This resulted in cases where the negative excited population was replaced with a value much greater than the NLTE ground pop resulting in extremely large partition function values which in some cases then overflowed the float limit. Additionally, negative NLTE ground populations were just replaced with the LTE value. However, if the ground population is negative this means the entire solution for that ion is unlikely to be sensible. If the NLTE ground population is negative now therefore revert to using LTE populations for the element. 
 -There were previously no checks on population inversions so have also added checks such that if the solver predicts a    large population inversion the level populations of the element are set to their LTE values - large population inversions can also lead to partition function values which overflow the float limit causing the code to fail. 

This pull request provides a temporary solution to problems that can occur when the NLTE solver does not produce a temporary solution. The plan is to add functionality in the future so that when the NLTE solver fails to a produce a sensible solution for an individual ion that specific ion can be set to LTE with the NLTE solver called again for the element with that ion excluded.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,10 +131,10 @@ jobs:
               working-directory: tests/${{ matrix.testname }}_testrun/
               run: cat job0/estimators*.out
 
-            - name: cat job0 output log
+            - name: cat job output logs
               if: always()
               working-directory: tests/${{ matrix.testname }}_testrun/
-              run: cat job0/output_0-0.txt
+              run: cat job0/output_*.txt
 
             - name: Checksum job0 output files
               if: always()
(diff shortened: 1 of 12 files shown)


fionntancallan:
@lukeshingles I think I've missed something with the syntax of find_uppermost_ion when trying to make it accessible outside of ltepop.cc, any ideas?

lukeshingles:
> @lukeshingles I think I've missed something with the syntax of find_uppermost_ion when trying to make it accessible outside of ltepop.cc, any ideas?

Definitions in the anonymous namespace block are not accessible outside the translation unit (*.cc file). You'll need to move it down past
```c++
} // anonymous namespace
```

lukeshingles:
Superseded by #187

PR #160, opened 2024-11-27
Ignore spurious gamma emission from nuclear de-excitations included in betaminusdecays.txt

lukeshingles:
These appear to be related to nuclear excited-state transitions, which are not handled by artis. Ignoring these gives good agreement with the total energy release rates (Qdot) from the network calculation.

Should not be significant for the published model, since (with uranium data fixed as it was) the abnormally high gamma emission is only seen after one day when gamma ray thermalisation is very inefficient. Will be more significant for higher-mass models.

beta_release_ratios_tot_Hotokezaka_Yeall_uraniumfix


lukeshingles:
Running this on the Just+2023 sym-n1-a6 model shows no significant difference. energy release is still exceeding Qdot and should be fixed.

PR #161, opened 2024-12-05
Add initial support for dielectronic recombination

ssim:
Partially addresses #219

This is for implementing dielectronic recombination into ARTIS.

Current status (5th Dec 2024):

- Draft infrastructure to store autoionization data for levels. 
- Autoionization and capture into resonances implemented into nlte population solver
- Preliminary testing for Si ions in 1 zone model appears to run and give roughly correct effect
- Pushing to GitHub to keep a safe copy! But this is not in a state yet to actually merge (so please don't anyone do that!)
- File attached here is the format proposed for supplying a list of autoionization processes. Format is pretty simple: list of Z upper_ion upper_level lower_ion lower_level A_auto

Still to do (5th Dec 2024):

- Need to figure out how to handle with superlevels.
- Likely don't actually want the autoionizing levels to take up space in the stored populations (i.e. should probably include them in the solver, but not include them in the stores). Not yet implemented.
- Lower priority: need to implement into the macro atom (that likely required some more infrastructure stuff added)
- Lower priority still: need to implement into the heating/cooling
- Need to properly test

**NOTE: I think that the superevel renorm factors currently in the code may be wrong for ionization processes. Need to check - but should discuss with Luke and see if we can fix. Don't think that actually will affect anything, but we should either fix or remove and abort.**

[autoion.txt](https://github.com/user-attachments/files/18027641/autoion.txt)

### DIFF
--- a/atomic.h
+++ b/atomic.h
@@ -317,9 +317,17 @@ inline auto get_includedions() -> int {
   return globals::elements[element].ions[ion].nlevels_excited_nlte;
 }
 
+// Returns the number of autoionising levels for an ion
+[[nodiscard]] inline auto get_nlevels_autoion(const int element, const int ion) -> int {
+  assert_testmodeonly(element < get_nelements());
+  assert_testmodeonly(ion < get_nions(element));
+  return globals::elements[element].ions[ion].nlevels_autoion;
+}
+
 // ion has NLTE levels, but this one is not NLTE => is in the superlevel
 [[nodiscard]] inline auto level_isinsuperlevel(const int element, const int ion, const int level) -> bool {
-  return (!is_nlte(element, ion, level) && level != 0 && (get_nlevels_excited_nlte(element, ion) > 0));
+  return (!is_nlte(element, ion, level) && level != 0 && (get_nlevels_excited_nlte(element, ion) > 0) &&
+          (level < (get_nlevels(element, ion) - get_nlevels_autoion(element, ion))));
 }
 
 [[nodiscard]] inline auto get_nlevels_groundterm(const int element, const int ion) -> int {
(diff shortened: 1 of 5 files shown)


lukeshingles:
The code generally looks good to me (including matching the style of the surrounding code!). Possibly I missed something, but it looks like you could avoid the two-pass reading of the autoion.txt and just append to the vector as lines are read. Something like:
```c++
globals::elements[element].ions[lowerion].levels[lowerlevel].allautoion_startdown = temp_allautoion.size();
temp_allautoion.push_back({.autoion_A = autoion_A, .elementindex = element, ...});
```
The nice thing about vectors is that they autoresize as items are appended. I only avoid doing that for very large data structures, because the allocated size can be bigger than the real size by up to a a factor of two. Since the vector is probably going to be fairly small, and it will be copied into precisely-sized shared memory before being erased anyway, that shouldn't be a problem.

clang-tidy is telling you that a couple of variables can be marked as constants. 

When autoion.txt is ready, you can commit a copy in the repository's data folder and we can make a test case to prevent breakage. The autoion.txt file would be slightly easier/faster to read in python if it followed single space delimiters instead of fixed width format, but if the file size is going to be small, this is not too important. Fixed width is also easier for hand editing.

(It's not possible to merge a draft pull request, so no risk of that!)

lukeshingles:
The branch is now up to date with develop. The previous state of the branch is preserved in the branch diel_devel_20250709.

The functionality has no test, so I can't say for sure whether this version works as it did before the rebase (perhaps you could do some quick manual tests?).

Since the new functionality only activates when autoion.txt is found and shouldn't have much effect on performance otherwise, I suggest we get this merged ASAP to avoid having to continually resolve merge conflicts.

lukeshingles (dismissed):

lukeshingles (dismissed):

lukeshingles (approved):

PR #162, opened 2024-12-20
Adding submit file for kelvin2 @ qub

fmcneill49:
This adds a submit script for Kelvin at qub: tested and working on kelvin on 20.12.24

### DIFF
--- a/scripts/artis-kelvin2.sh
+++ b/scripts/artis-kelvin2.sh
@@ -1,8 +1,15 @@
 #!/bin/bash -x
+<<<<<<< HEAD
+#SBATCH --ntasks=1024
+#SBATCH --ntasks-per-node=128
+#SBATCH --time=24:00:00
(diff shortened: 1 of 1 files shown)


scripts/artis-kelvin2.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -x
+#SBATCH --ntasks=128
+#SBATCH --ntasks-per-node=128
+#SBATCH --time=00:30:00
```
lukeshingles, line 4: We usually aim for about ~1000 cores and 24 hours by default with these job scripts. Is this possible on Kelvin?

lukeshingles:
Thanks for the contribution. Will merge it after we answer the cores/time question.

lukeshingles:
Sorry, didn't realise you'd updated this (I only get notified of comments, not commits). Let's merge.

PR #186, opened 2025-04-18
Fix DO_TITER option

AlexHls:
In the current ``develop`` branch, enabling the ``DO_TITER`` does not seem to work. Superficially, this is a compile issue, but I'm not sure if there are some deeper issues.

So far I've been able to fix the compile issues and get it to run with 3 iterations for timesteps after the first timestep (i.e. setting ``globals::n_titer = (globals::timestep > 1) ? 3 : 1;``), but my test setup crashes later for some other reason I haven't been able to figure out. Somebody more knowledgable than me may need to run some verification tests to make sure this works as intended.

### DIFF
--- a/input.cc
+++ b/input.cc
@@ -1619,7 +1619,7 @@ void read_atomicdata() {
 
 // read input.txt, atomic data, and ejecta model
 void input(int rank) {
-  globals::n_titer = 1;
+  globals::n_titer = (globals::timestep < -1) ? 3 : 1;
   globals::lte_iteration = false;
 
   printout("[info] input: do n_titer %d iterations per timestep\n", globals::n_titer);
--- a/radfield.cc
+++ b/radfield.cc
@@ -636,8 +636,8 @@ void initialise_prev_titer_photoionestimators() {
 #ifdef DO_TITER
   std::ranges::fill(globals::ffheatingestimator_save, -1.);
   std::ranges::fill(globals::colheatingestimator_save, -1.);
-  std::ranges::fill(globals::J_reduced_save, -1.);
-  std::ranges::fill(globals::nuJ_reduced_save, -1.);
+  std::ranges::fill(J_reduced_save, -1.);
+  std::ranges::fill(nuJ_reduced_save, -1.);
   for (int nonemptymgi = 0; nonemptymgi < grid::get_nonempty_npts_model(); nonemptymgi++) {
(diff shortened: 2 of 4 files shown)


lukeshingles:
I probably won’t get to this properly until April 28, but can you tell me more about why you need this feature? If there’s a need for it, we should add a test to the CI to ensure it stays in a working state. I personally only used it for a one-off experiment many years ago.

AlexHls:
We're currently running some experiments including a pure H/He envelope. In those cases the ionization stage seems to jump in between timesteps (in case of H from mostly neutral to fully ionizes), causing issues in e.g. the electron temperature etc. @ssim suggested that using multiple iterations per timestep may help to converge the ionization stage (given that artis lacks the infrastructure to limit the maximum change in ionization).

Assuming that this solves our issues, I would have a more long-term need for this feature as several of my models would involve such envelopes.

lukeshingles:
Ok! If this is going to be useful, I suggest we add another artisoptions.h function, something like:
```c++
constexpr int NUM_TIMESTEP_ITERATIONS(int timestep) { return (timestep <= -1) ? 3 : 1; }
```
Then this can be enabled for one of the CI tests (maybe in classicmode_1d) and we'll store the checksums. Hopefully you're using classic options? The timestep iteration stuff isn't implemented for the binned radiation field or the detailed bound-free estimators that we use for NLTE calculations.

AlexHls:
> Hopefully you're using classic options? The timestep iteration stuff isn't implemented for the binned radiation field or the detailed bound-free estimators that we use for NLTE calculations.

Ah, this explains why it runs in my test setup (classic), but not in my production setup, which runs on NLTE settings (or at least a binned radiation field) since we're mostly interested in nebular spectra. Is this something that could be implemented rather easily or would this require some major code changes? I'm willing to do most of the heavy lifting here, at least as far as my expertise goes.

lukeshingles:
Yes, I think you could do it with a bit of work. If you happen to come up with a neat way to do this (i.e. not just duplicating all of the arrays with a ‘_save’ suffix) that would be even better, but even the current way extended to those extra features would be fine to merge.

lukeshingles (approved):

lukeshingles:
I'll merge this compilation fix now. If you want to do any further work on the TITER mode, that can be done in a separate PR.

PR #187, opened 2025-04-28
Allow removing unimportant ion stages from the NLTE population solver

fionntancallan:
Add functionality that allows the range of ions that are used in the NLTE solution to be dynamically changed if the NLTE solution fails due to either negative populations or large population inversions:

-Previously when the NLTE population was negative for a level it was directly replaced with the LTE level population. This resulted in cases where the negative excited population was replaced with a value much greater than the NLTE ground pop resulting in extremely large partition function values which in some cases then overflowed the float limit resulting in the simulation failing. Additionally, negative NLTE ground populations were just replaced with the LTE value. However, if the ground population is negative this means the entire solution for that ion is unlikely to be sensible.

-There were previously no checks on population inversions. Large population inversions can lead to partition function values which overflow the float limit causing the simulation to fail. 

The functionality added in this pull request should solve issues related to the NLTE solver giving negative populations or large populations inversions for ions which have very low populations (e.g for neutral ion states when the dominant ionisation state is a highly ionised state and vice versa) by allowing the NLTE solution to be recalculated when the solver fails but with the individual ion that is causing the issues excluded from the solution for that element. Previously the NLTE solution could only be calculated including all ions included in the simulation for a given element and there was only functionality to set the populations of the entire element to LTE values of the NLTE solver failed.


fionntancallan:
@lukeshingles, any ideas on why this is failing with a "terminate called after throwing an instance of 'std::bad_array_new_length'"? All I have done is add the nions_used and first_ion_used variables and set them equal to what was previously used in the get_element_nlte_dimension and get_max_nlte_dimension functions. 

Is there a way to run the tests locally so that I can get more detailed information to diagnose this?

lukeshingles:
@fionntancallan the most common cause of that would be trying to resize the vector to a negative length. You can look in the ci.yml workflow and run those same commands locally.

fionntancallan:
@lukeshingles in "solve_nlte_pops_element" when assigning memory to the rate matrices, e.g.:

    resize_exactly(vec_rate_matrix_rad_bb, max_nlte_dimension * max_nlte_dimension);
    rate_matrix_rad_bb = gsl_matrix_view_array(vec_rate_matrix_rad_bb.data(), nlte_dimension, nlte_dimension).matrix;
    gsl_matrix_set_all(&rate_matrix_rad_bb, 0.);
    
Why is max_nlte_dimension used for the resize? Since solve_nlte_pops_element is only called for one element at a time could you not just use nlte_dimension for the resize making max_nlte_dimension redundant? Or am I missing something?

fionntancallan:
Is it because the expensive operation is when you have to resize to a matrix bigger than the memory allocated for the previous iteration of the matrix, so its generally better to allocate the max memory and then reduce what is actually utilised with the gsl_matrix_view_array?

lukeshingles:
> Is it because the expensive operation is when you have to resize to a matrix bigger than the memory allocated for the previous iteration of the matrix, so its generally better to allocate the max memory and then reduce what is actually utilised with the gsl_matrix_view_array?

Yes, memory allocations and deallocations are slow and they cause memory fragmentation. It's better to allocate just once if it's possible to know ahead of time how much memory is required. I don't see why you would have to touch that code though, since you're only potentially reducing the size of the matrix, the max_nlte_dimension should not change.

fionntancallan:
> > Is it because the expensive operation is when you have to resize to a matrix bigger than the memory allocated for the previous iteration of the matrix, so its generally better to allocate the max memory and then reduce what is actually utilised with the gsl_matrix_view_array?
> 
> Yes, memory allocations and deallocations are slow and they cause memory fragmentation. It's better to allocate just once if it's possible to know ahead of time how much memory is required. I don't see why you would have to touch that code though, since you're only potentially reducing the size of the matrix, the max_nlte_dimension should not change.

Ok thanks, the output of max nlte_dimension shouldn't change but it works by looping through all the elements and calling get_element_nlte_dimension for each element which now needs two extra arguments which can be applied if the number of ions have been reduced. This change in dimension shouldn't apply to the call of get_element_nlte_dimension from get_max_nlte_dimension provided the arguments passed are consistent with the entire ion being treated in NLTE

fionntancallan:
@lukeshingles do you prefer this branch is brought up to date with the current develop branch with update with rebase or update with merge commit?

lukeshingles:
> @lukeshingles do you prefer this branch is brought up to date with the current develop branch with update with rebase or update with merge commit?   

Either one is good since it will look the same once it is merged anyway.

fionntancallan:
@lukeshingles the new functionality seems to be working for single zone test simulations. What is the best way to proceed from here. Are there larger scale tests you run to make sure everything is behaving? The new functionality changes the results of the automated tests so its probably a good idea to do some more comprehensive testing before we would update the tests.

nltepop.cc
```diff
@@ -372,23 +377,33 @@ auto get_element_superlevelpartfuncs(const int nonemptymgi, const int element) -
 }
 
 // get the maximum NLTE dimension for any of the included elements
-[[nodiscard]] auto get_max_nlte_dimension() {
+[[nodiscard]] auto get_max_nlte_dimension(int nions_used, int first_ion_used) -> int {
```
lukeshingles: These parameters are not being used and should be removed.

lukeshingles:
> @lukeshingles the new functionality seems to be working for single zone test simulations. What is the best way to proceed from here. Are there larger scale tests you run to make sure everything is behaving? The new functionality changes the results of the automated tests so its probably a good idea to do some more comprehensive testing before we would update the tests.

Please add some description of your new options to artisoptions_doc.md. Currently, our two NLTE tests do not pass through the NLTE error condition and recovery steps. I suggest duplicating and modifying input model/time range for the nebular_1d_3dgrid test so that it gets into this kind of state and preserves the functionality of your new options.

nltepop.cc
```diff
@@ -39,24 +39,25 @@ constexpr bool individual_process_matrices = true;
 
 // this is the index for the NLTE solver that is handling all ions of a single element
 // This is NOT an index into grid::modelgrid[nonemptymgi].nlte_pops that contains all elements
-auto get_nlte_vector_index(const int element, const int ion, const int level) -> int {
+auto get_nlte_vector_index(const int element, const int ion, const int level, const int first_ion_used) -> int {
```
lukeshingles, line 42: Add
```c++
assert_testmodeonly(first_ion_used >= 0);
assert_testmodeonly(first_ion_used < get_nions(element));
```

nltepop.cc
```diff
@@ -354,10 +358,11 @@ auto get_element_superlevelpartfuncs(const int nonemptymgi, const int element) -
   return superlevel_partfuncs;
 }
 
-[[nodiscard]] auto get_element_nlte_dimension(const int element) -> int {
+[[nodiscard]] auto get_element_nlte_dimension(const int element, const int nions_used, const int first_ion_used)
+    -> int {
   int nlte_dimension = 0;
-  const int nions = get_nions(element);
-  for (int ion = 0; ion < nions; ion++) {
+  // const int nions_prev = get_nions(element);
```
lukeshingles: Remove commented code.

nltepop.cc
```diff
@@ -354,10 +358,11 @@ auto get_element_superlevelpartfuncs(const int nonemptymgi, const int element) -
   return superlevel_partfuncs;
 }
 
-[[nodiscard]] auto get_element_nlte_dimension(const int element) -> int {
+[[nodiscard]] auto get_element_nlte_dimension(const int element, const int nions_used, const int first_ion_used)
+    -> int {
```
lukeshingles, line 361:
```c++
assert_testmode(nions_used >= 0);
assert_testmode(nions_used < get_nions(elements));
assert_testmode(first_ion_used >= 0);
assert_testmode(first_ion_used < get_nions(element));
assert_testmode((first_ion_used + nions_used - 1) < get_nions(element));
```

artisoptions_christinenonthermal.h
```diff
@@ -89,6 +89,18 @@ constexpr bool USE_LUT_PHOTOION = false;
...
 
+constexpr bool STRICT_POPULATION_CHECKING = false;
+
+constexpr bool NLTE_LIMIT_ION_STAGES_AFTER_FAILURE = false;
+
+constexpr float POP_INVERSION_FACTOR_SOLVER_FAIL = 1000.;
+
+constexpr float POP_INVERSION_FACTOR_SMALL_INVERSION_WARNING = 10.;
+
+constexpr float POPULATION_CUT_REMOVE_ION = 100.;
```
lukeshingles: If these options are going to have exactly the same values in every one of our presets, then think about whether we really need to be adding them to artistoptions.h at all, versus just placing the number directly in nltepop.cc. (I realise that I haven't followed my own advice here with e.g., NT_MAX_AUGER_ELECTRONS, but hindsight is 20/20 and I plan to remove these kind of unneeded options in the future).

If you think we should keep them, then could you please use a prefix like STRICT_POPULATION_CHECKING_INVERSION_FACTOR_SOLVER_FAIL just to make it clear that the option has no effect if STRICT_POPULATION_CHECKING is false.

I would also beware of placing limits that have a dependence on the absolute values of the density and element populations (POPULATION_CUT_REMOVE_ION). These limiting conditions should be ratios of internal quantities like level populations per [element, ion, or level] population as you have done with the POP_INVERSION_FACTOR options.

fionntancallan: They have all the same values in the presets at the minute because those are the values I have been using in my testing so far but they are definitely something that I am likely to play with in the future. For example with the new dynamic ion ranges in the NLTE solver we may be able to be much more aggressive on what factor of population inversion results in a solver fail but I haven't ran enough simulations for different models to determine exactly what the optimal number would be. If I converge on what seem to be sensible fixed values for these numbers in the future they could be removed as options but for now I would prefer to have the flexibility to change them easily in the artisoptions.h 

I agree POPULATION_CUT_ION should instead be a ratio instead - probably its should be the ratio of the ion population to the total element population

nltepop.cc
```diff
@@ -186,10 +187,11 @@ void print_element_rates_summary(const int element, const int modelgridindex, co
                                  const gsl_vector *popvec, const gsl_matrix *rate_matrix_rad_bb,
                                  const gsl_matrix *rate_matrix_coll_bb, const gsl_matrix *rate_matrix_ntcoll_bb,
                                  const gsl_matrix *rate_matrix_rad_bf, const gsl_matrix *rate_matrix_coll_bf,
-                                 const gsl_matrix *rate_matrix_ntcoll_bf) {
+                                 const gsl_matrix *rate_matrix_ntcoll_bf, const int first_ion_used,
+                                 const int nions_used) {
   const auto nonemptymgi = grid::get_nonemptymgi_of_mgi(modelgridindex);
-  const int nions = get_nions(element);
-  for (int ion = 0; ion < nions; ion++) {
+  // const int nions = get_nions(element);
```
lukeshingles: Remove commented code.

nltepop.cc
```diff
@@ -531,37 +543,74 @@ void nltepop_matrix_add_ionisation(const int nonemptymgi, const int element, con
...
+            "ions "
+            "%g\n",
+            get_ionstage(element, ion), get_ionstage(element, upperion), Y_nt_thisupperion);
+        Y_nt_thisupperion += Y_nt * nonthermal::nt_ionization_upperion_probability(
+                                        nonemptymgi, element, ion, upperion_outside_NT_ionisation_range, false);
+        printout("Y_nt_thisupperion contribution from ionstage=%d (above upperion=%d) = %g\n",
+                 get_ionstage(element, upperion_outside_NT_ionisation_range), get_ionstage(element, upperion),
+                 Y_nt * nonthermal::nt_ionization_upperion_probability(nonemptymgi, element, ion,
+                                                                       upperion_outside_NT_ionisation_range, false));
+        printout("Y_nt_thisupperion for upperion top allowed=%d after adding NT ionisation for stripped ion %g\n",
```
lukeshingles: Unused debugging code?

nltepop.cc
```diff
@@ -531,37 +543,74 @@ void nltepop_matrix_add_ionisation(const int nonemptymgi, const int element, con
...
 
         atomicadd(*gsl_matrix_ptr(rate_matrix_ntcoll_bf, lower_index, lower_index),
                   -Y_nt_thisupperion * s_renorm[level]);
         atomicadd(*gsl_matrix_ptr(rate_matrix_ntcoll_bf, upper_groundstate_index, lower_index),
                   Y_nt_thisupperion * s_renorm[level]);
       }
     }
+    if (nonthermal::nt_ionisation_maxupperion(element, ion) > (first_ion_used + nions_used - 1)) {
+      Y_nt_thisupperion_sum += Y_nt_thisupperion;
+      printout("Y_nt = %g Y_nt_thisupperion_sum = %g\n", Y_nt, Y_nt_thisupperion_sum);
```
lukeshingles: Unused debugging code?

tests/setup_nltephotospheric_dynamic_ion_range_1d_1dgrid.sh
```diff
@@ -0,0 +1,52 @@
...
+
+cd $runfolder
+
+sed -i'' -e 's/constexpr int MPKTS.*/constexpr int MPKTS = 250;/g' artisoptions.h
+
+sed -i'' -e 's/constexpr auto GRID_TYPE.*/constexpr auto GRID_TYPE = GridType::SPHERICAL1D;/g' artisoptions.h
+
+sed -i'' -e 's/constexpr int CUBOID_NCOORDGRID_X.*/constexpr int CUBOID_NCOORDGRID_X = 50;/g' artisoptions.h
+sed -i'' -e 's/constexpr int CUBOID_NCOORDGRID_Y.*/constexpr int CUBOID_NCOORDGRID_Y = 50;/g' artisoptions.h
+sed -i'' -e 's/constexpr int CUBOID_NCOORDGRID_Z.*/constexpr int CUBOID_NCOORDGRID_Z = 50;/g' artisoptions.h
```
lukeshingles: These lines can be removed since they don't apply with a 1D spherical grid.

nltepop.cc
```diff
@@ -795,184 +914,240 @@ void solve_nlte_pops_element(const int element, const int nonemptymgi, const int
...
+            "successfully solved NLTE matrix when reducing ions used for element to Z=%d ionstage=%d to ionstage=%d\n",
+            atomic_number, get_ionstage(element, first_ion_used),
+            get_ionstage(element, first_ion_used + nions_used - 1));
       }
 
-      gsl_vector_set(&balance_vector, get_nlte_vector_index(element, ion, index_ion_ground), nnion);
+    } else {
+      retry_with_fewer_ions = true;
+      if (gsl_vector_get(&popvec, get_nlte_vector_index(element, first_ion_used + nions_used - 1, 0, first_ion_used)) <
+          POPULATION_CUT_REMOVE_ION) {
```
fionntancallan: @lukeshingles currently this is checking whether the ground population of the ion which caused the solver fail has an absolute population of less than 100 if the ion is cut. If I want to instead make the check a ratio of the ion population to the total element population is there an easier/better way to get the ion population other than looping through the population vector and summing the populations of all the individual levels for that ion? I suppose I could instead assume that the majority of the population for the ion is in the ground state - but this may not be consistent with cases when the solver fails because of large population inversions reletive to the ground population.

lukeshingles: The ion populations get calculated at some point in the solver. If it’s not updated already here, then you might need to move it up so you can use get_nnion()

fionntancallan:
@lukeshingles I want to run some full simulations to test the new functionality in more detail and see if it changes anything in terms of the synthetic observables predicted. Are there specific runs on the ARTIS google drive which you generally rerun after an update to the code?

lukeshingles:
> @lukeshingles I want to run some full simulations to test the new functionality in more detail and see if it changes anything in terms of the synthetic observables predicted. Are there specific runs on the ARTIS google drive which you generally rerun after an update to the code?

I think we discussed re-running the W7 and a Shen2018 with your new options enabled to see if it affects the nebular spectra (hopefully not). Google Drive path: `ARTIS/artis_runs_published/shinglesetal2022`

But you don't necessarily need to do that before we merge the PR. You can always follow up with another PR if you want to make further changes to your new experimental options.

tests/nltephotospheric_dynamic_ion_range_1d_1dgrid_inputfiles/recombrates.txt
lukeshingles, line 1: If any of input files in your new test are the same as those in nebular_1d_grid, then just copy them from there instead of duplicating these files and taking up more space in the repository. Look at how nebular_1d_3dgrid_limitbfest test copies its input files from nebular_1d_grid

tests/nebular_dynamic_ion_range_1d_1dgrid_inputfiles/abundances.txt
lukeshingles, line 1: This file is not referenced anywhere.

artisoptions_christinenonthermal.h
```diff
@@ -89,6 +89,18 @@ constexpr bool USE_LUT_PHOTOION = false;
...
+
+constexpr bool NLTE_LIMIT_ION_STAGES_AFTER_FAILURE = false;
+
+constexpr float STRICT_POPULATION_CHECKING_INVERSION_FACTOR_SOLVER_FAIL = 1000.;
+
+constexpr float STRICT_POPULATION_CHECKING_INVERSION_FACTOR_PRINTOUT_WARNING = 10.;
+
+constexpr double LEVEL_POPULATION_RATIO_WITH_ELEMENT_ALLOW_REMOVE_ION = 1e-9;
+
+constexpr bool RECALL_FIND_UPPERMOST_ION_WHEN_SETTING_ELEMENT_POPS_LTE = false;
```
lukeshingles: Please make it clear that this only affects the NLTE pop solver when there is a failure. e.g.
NLTEPOP_FAILURE_USE_FIND_UPPERMOST_ION_FOR_LTE_RESET

Functionality can change down the line quite easily, but renaming options is something that we should try to avoid, since it makes it very difficult to apply the same set of options to different artis versions. Better to get it right from the start, if possible.

artisoptions_christinenonthermal.h
```diff
@@ -89,6 +89,18 @@ constexpr bool USE_LUT_PHOTOION = false;
...
 
+constexpr bool STRICT_POPULATION_CHECKING = false;
+
+constexpr bool NLTE_LIMIT_ION_STAGES_AFTER_FAILURE = false;
+
+constexpr float STRICT_POPULATION_CHECKING_INVERSION_FACTOR_SOLVER_FAIL = 1000.;
+
+constexpr float STRICT_POPULATION_CHECKING_INVERSION_FACTOR_PRINTOUT_WARNING = 10.;
+
+constexpr double LEVEL_POPULATION_RATIO_WITH_ELEMENT_ALLOW_REMOVE_ION = 1e-9;
```
lukeshingles: Does this do anything if NLTE_LIMIT_ION_STAGES_AFTER_FAILURE is false? if not, make it clear, e.g.,

NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION

nltepop.cc
```diff
@@ -795,183 +893,360 @@ void solve_nlte_pops_element(const int element, const int nonemptymgi, const int
...
+      // to this, then check the excited populations, then check the superlevel population if it exists.
+      // Check all of these relative to a factor of 1e6 or 1e9 of the total element population.
+      // Do this for each the top ion and the bottom ion here then set flags for each of these e.g.
+      // top_ion_can_cut_from_solution = true; If looping through each of these with an if condition is too slow
+      // then instead could so the sum with an fabs(pop) for each population value so that negative populations
+      // don't cancel out positive populations and make it seem like the population of an ion is very low when
+      // actually it is just a big negative population cancelling out the larger positive populations.
+
+      bool remove_top_ion_from_solution = true;
+      bool remove_bottom_ion_from_solution = true;
```
lukeshingles, line 1088: If I'm reading this correctly, this code path runs even when NLTE_LIMIT_ION_STAGES_AFTER_FAILURE is false. Am I mistaken?

nltepop.cc
```diff
@@ -795,183 +893,360 @@ void solve_nlte_pops_element(const int element, const int nonemptymgi, const int
...
+              "(%g/%g) > (%g) LEVEL_POPULATION_RATIO_WITH_ELEMENT_ALLOW_REMOVE_ION)\n",
+              superlevel_pop_top_ion, nnelement, LEVEL_POPULATION_RATIO_WITH_ELEMENT_ALLOW_REMOVE_ION);
+          remove_top_ion_from_solution = false;
+        }
+      }
+
+      if (remove_top_ion_from_solution) {
+        printout(
+            "  Passed checks: removing top ion from NLTE solution and attempting to re-solve rate (ionstage %d, ground "
+            "pop %g, fabs(excited pops) sum %g, superlevel pop %g, nnelement %g) matrix\n",
```
lukeshingles: Is this code definitely unreachable when NLTE_LIMIT_ION_STAGES_AFTER_FAILURE is false?

fionntancallan:
> > @lukeshingles I want to run some full simulations to test the new functionality in more detail and see if it changes anything in terms of the synthetic observables predicted. Are there specific runs on the ARTIS google drive which you generally rerun after an update to the code?
> 
> I think we discussed re-running the W7 and a Shen2018 with your new options enabled to see if it affects the nebular spectra (hopefully not). Google Drive path: `ARTIS/artis_runs_published/shinglesetal2022`
> 
> But you don't necessarily need to do that before we merge the PR. You can always follow up with another PR if you want to make further changes to your new experimental options.

Yes I can run those after we merge the PR,I was just wondering what the exact versions to compare to are since there are a few different versions of the W7 and Shen model in that folder. I guess its the subchdet_shen2018_1.00_5050_20211111_150_410d_ntfixnewphixsbfest_2e9pkt_virgo and w7_outercut_20211111_150_410d_ntfixnewphixsbfest_ntworkfn_2e9pkt_virgo runs I should be comparing to but wanted to double check

tests/nebular_dynamic_ion_range_1d_1dgrid_inputfiles/abundances.txt
fionntancallan: I can't see that file or folder anywhere in my most recent version?

lukeshingles:
> > > @lukeshingles I want to run some full simulations to test the new functionality in more detail and see if it changes anything in terms of the synthetic observables predicted. Are there specific runs on the ARTIS google drive which you generally rerun after an update to the code?
> > 
> > 
> > I think we discussed re-running the W7 and a Shen2018 with your new options enabled to see if it affects the nebular spectra (hopefully not). Google Drive path: `ARTIS/artis_runs_published/shinglesetal2022`
> > But you don't necessarily need to do that before we merge the PR. You can always follow up with another PR if you want to make further changes to your new experimental options.
> 
> Yes I can run those after we merge the PR,I was just wondering what the exact versions to compare to are since there are a few different versions of the W7 and Shen model in that folder. I guess its the subchdet_shen2018_1.00_5050_20211111_150_410d_ntfixnewphixsbfest_2e9pkt_virgo and w7_outercut_20211111_150_410d_ntfixnewphixsbfest_ntworkfn_2e9pkt_virgo runs I should be comparing to but wanted to double check

Yes for the subchdet, but for w7 use w7_outercut_20210331_150_410d_2e9pkt_fixntfrac_juwels.

The artis options have evolved a bit since there, but you should be able to simply take the latest artisoptions_nltenebular.h and edit the packet count. 2e9 packets is probably overkill, and 1e9 packets should look nearly identical anyway. So for 960 cores, set MPKTS to 1042000.

nltepop.cc
fionntancallan: Good catch, I've added an if statement around this code so that it is only accessed when NLTE_LIMIT_ION_STAGES_AFTER_FAILURE==True. Let me know if you are happy with this solution or would prefer we do something else here

nltepop.cc
fionntancallan: See comment above

scripts/artis-cosma8.sh
```diff
@@ -4,10 +4,11 @@
...
 ##SBATCH --ntasks 1920
 ##SBATCH --ntasks 3072
-#SBATCH --time=70:00:00
-#SBATCH --partition=cosma8-milan
-#SBATCH --account=dp033
+#SBATCH -J artis
+#SBATCH -p cosma8
+#SBATCH -A dp033
 #SBATCH --exclusive
+#SBATCH -t 70:00:00
```
lukeshingles: What is the reason for these changes?

nltepop.cc
```diff
@@ -473,24 +484,24 @@ void nltepop_matrix_add_boundbound(const int nonemptymgi, const int element, con
 
 void nltepop_matrix_add_ionisation(const int nonemptymgi, const int element, const int ion,
                                    const std::vector<double> &s_renorm, gsl_matrix *rate_matrix_rad_bf,
-                                   gsl_matrix *rate_matrix_coll_bf) {
-  assert_always((ion + 1) < get_nions(element));  // can't ionise the top ion
+                                   gsl_matrix *rate_matrix_coll_bf, const int first_ion_used, const int nions_used) {
+  assert_always((ion + 1) < nions_used + first_ion_used);  // updated to cover case when ions have been stripped
```
lukeshingles: Not a bug, but please add brackets around (nions_used + first_ion_used) to make the order of operations clear. The old comment "can't ionise the top ion" made a lot more sense to me as a hint about what the condition is actually testing. The history of changes should be in git, not code comments.

nltepop.cc
```diff
@@ -531,25 +542,43 @@ void nltepop_matrix_add_ionisation(const int nonemptymgi, const int element, con
 }
 
 void nltepop_matrix_add_nt_ionisation(const int nonemptymgi, const int element, const int ion,
-                                      const std::vector<double> &s_renorm, gsl_matrix *rate_matrix_ntcoll_bf) {
+                                      const std::vector<double> &s_renorm, gsl_matrix *rate_matrix_ntcoll_bf,
+                                      const int first_ion_used, const int nions_used) {
   // collisional ionization by non-thermal electrons
 
-  assert_always(ion + 1 < get_nions(element));  // can't ionise the top ion
+  assert_always(ion + 1 < nions_used + first_ion_used);  // updated to cover case when ions have been stripped
```
lukeshingles: As above, the old comment made more sense to me (and brackets would be helpful, although this was an issue with the existing code). If you think the new comment is better, please explain why.

nltepop.cc
```diff
@@ -531,25 +542,43 @@ void nltepop_matrix_add_ionisation(const int nonemptymgi, const int element, con
...
+    // solution so this contribution to the NT ionsiation rate isn't lost
+    if (nonthermal::nt_ionisation_maxupperion(element, ion) > (first_ion_used + nions_used - 1) &&
+        upperion == first_ion_used + nions_used - 1) {
+      for (int upperion_outside_NT_ionisation_range = first_ion_used + nions_used;
+           upperion_outside_NT_ionisation_range <= nonthermal::nt_ionisation_maxupperion(element, ion);
+           upperion_outside_NT_ionisation_range++) {
+        Y_nt_thisupperion += Y_nt * nonthermal::nt_ionization_upperion_probability(
+                                        nonemptymgi, element, ion, upperion_outside_NT_ionisation_range, false);
+      }
+    }
```
lukeshingles: I'm a bit confused about the logic here. Couldn't you keep the non-thermal code all as it was and simply change:
```c++
const int upper_groundstate_index = get_nlte_vector_index(element, upperion, 0, first_ion_used);
```
to:
```c++
const int upperion_clamped = std::min(upperion, first_ion_used + nions_used - 1);
const int upper_groundstate_index = get_nlte_vector_index(element, upperion_clamped, 0, first_ion_used);
```

nltepop.cc
```diff
@@ -730,27 +769,86 @@ void set_element_pops_lte(const int nonemptymgi, const int element) {
...
   for (size_t row = 0; row < nlte_dimension; row++) {
     double recovered_balance_vector_elem = 0.;
     gsl_vector_const_view row_view = gsl_matrix_const_row(rate_matrix, row);
     gsl_blas_ddot(&row_view.vector, &x, &recovered_balance_vector_elem);
 
-    const auto [ion, level] = get_ion_level_of_nlte_vector_index(row, element);
-
+    const auto [ion, level] = get_ion_level_of_nlte_vector_index(row, element, first_ion_used, nions_used);
+    if (level == 0) {
+      row_ground_state = row;
```
lukeshingles: This does work, but I think it would better to be consistent what what you have further down:
```c++
const int index_ion_ground = get_nlte_vector_index(element, ion, 0, first_ion_used);
```
With a constant, it's more clear exactly what this variable contains and whether it has been set correctly.

scripts/artis-cosma8.sh
fionntancallan: I think there was a conflict when I tried to merge the changes from develop into the branch. Happy to change it back but it looks like the only functional change is the partition being requested which I suppose we want to change to cosma8 anyway?

lukeshingles: Please revert the changes for this PR.

tests/nebular_dynamic_ion_range_1d_1dgrid_inputfiles/abundances.txt
lukeshingles, line 1: Is this file included by mistake? I don't see it referenced anywhere.

nltepop.cc
```diff
@@ -730,27 +769,86 @@ void set_element_pops_lte(const int nonemptymgi, const int element) {
...
     //          gsl_vector_get(x, row) / gsl_vector_get(x,get_nlte_vector_index(element,ion,0)));
 
-    if (gsl_vector_get(popvec, row) < 0.0) {
-      printout(
-          "  WARNING: NLTE solver gave negative population to index %zu (Z=%d ionstage %d level %d), pop = %g. "
-          "Replacing with LTE pop of %g\n",
-          row, get_atomicnumber(element), get_ionstage(element, ion), level,
-          gsl_vector_get(&x, row) * gsl_vector_get(pop_normfactor_vec, row), gsl_vector_get(pop_normfactor_vec, row));
-      gsl_vector_set(popvec, row, gsl_vector_get(pop_normfactor_vec, row));
+    if (STRICT_POPULATION_CHECKING) {
```
lukeshingles, line 789: Are you sure that you preserved the old negative population check when STRICT_POPULATION_CHECKING is false?

nltepop.cc
```diff
@@ -730,27 +769,86 @@ void set_element_pops_lte(const int nonemptymgi, const int element) {
...
+
+        if (gsl_vector_get(popvec, row_ground_state) * STRICT_POPULATION_CHECKING_INVERSION_FACTOR_SOLVER_FAIL <
+            (stat_weight(element, ion, 0) / stat_weight(element, ion, level)) * gsl_vector_get(popvec, row)) {
+          printout("large pop inversion - return matrix solve fail\n");
+          return false;
+        }
+        printout("relatively small pop inversion continue with NLTE solution\n");
+      }
+    } else {
+      if (gsl_vector_get(popvec, row) < 0.0) {
```
lukeshingles: } else if (..) {

nltepop.cc
fionntancallan: If I am understanding that correctly then the way you have suggested is definitely a cleaner way of doing this. Just to confirm the first atomicadd reduces the population of the lower level then the second atomicadd adds this to the ground of the upper ion? For the way you have suggested then if the upper ion is greater than the ion range of the solver the population is instead added to the top ion allowed by the solver?

lukeshingles: Yes. Is that not the how you think it should work? This already what we do in the non-thermal solver when we're limited by which ions have been included. For example, we can't multiply-ionise the second top ion, so we just redirect all of non-thermal ionisation rate to end up in the top ion.

fionntancallan: Yes that is how I intended it to work, I just wanted to double check that is what was happening. I will revert the changes I had previously made and add the upperion_clamped line instead as you suggested.

lukeshingles (approved):
Nice work. Let's merge it.

PR #199, opened 2025-06-11
Handle species with zero bound electrons (e.g. He III)

fionntancallan:


nonthermal.cc
```diff
@@ -656,6 +658,9 @@ void read_collion_data() {
 
         const int ioncharge = ionstage - 1;
         const int nbound = Z - ioncharge;  // number of bound electrons
+        if (nbound <= 0) {
+          continue;
+        }
```
lukeshingles: These lines could be moved to before 651 (partly to avoid a nonsensical log message)

lukeshingles:
Can you confirm that the zero-bound electron species are handled as you would expect in the log messages of the Spencer-Fano solver? (i.e. we're not printing out junk numbers for it)

nonthermal.cc
fionntancallan: I don't think this can be moved above line 651 because the ioncharge and therefore nbound come from the element that is being looped through starting from line 652. Do you mean line 661 to avoid the "No collisional ionisation data for Z=%d ionstage %d. Using Lotz approximation with ionpot" printout?

lukeshingles: I'm not sure how you're getting those line numbers, because they don't seem to match up with either the before or after line numbers on GitHub. The nbound check (and whatever necessary variables like ioncharge) can be the first thing in the ion loop that starts on line 647 before, or line 649 on the current commit of this PR. That avoids the log message stating that we're going to use the Lotz approximation for these ions, which is not true.

fionntancallan:
> Can you confirm that the zero-bound electron species are handled as you would expect in the log messages of the Spencer-Fano solver? (i.e. we're not printing out junk numbers for it)

I have been running the tests locally with He included in the atomic data set for this and noticed that for the current model used in the photospheric test there is only a single cell that is treated in NLTE because the grey cut of cells with tau>8 is very low for the photospheric phase. I suppose this helps speed up the tests so this might not be something we want to change. However in this cell the He fraction is so low that it doesn't get included when setting up the Spencer-fano solution. The simplest change I could make to the test which shouldn't alter the run time of the tests is to just increase the He fraction in the cell that is treated in NLTE in the tests. Just wanted to check if you think this is the best approach or would you recommend something else?

lukeshingles:
> > Can you confirm that the zero-bound electron species are handled as you would expect in the log messages of the Spencer-Fano solver? (i.e. we're not printing out junk numbers for it)
> 
> I have been running the tests locally with He included in the atomic data set for this and noticed that for the current model used in the photospheric test there is only a single cell that is treated in NLTE because the grey cut of cells with tau>8 is very low for the photospheric phase. I suppose this helps speed up the tests so this might not be something we want to change. However in this cell the He fraction is so low that it doesn't get included when setting up the Spencer-fano solution. The simplest change I could make to the test which shouldn't alter the run time of the tests is to just increase the He fraction in the cell that is treated in NLTE in the tests. Just wanted to check if you think this is the best approach or would you recommend something else?

Can you get something working here, we'll merge it, and then do a separate PR that fixes the test with no code changes? If the checksums need to change, then we don't want to be confused about exactly what caused the change (e.g. unrelated code in the diff).

fionntancallan:
> > > Can you confirm that the zero-bound electron species are handled as you would expect in the log messages of the Spencer-Fano solver? (i.e. we're not printing out junk numbers for it)
> > 
> > 
> > I have been running the tests locally with He included in the atomic data set for this and noticed that for the current model used in the photospheric test there is only a single cell that is treated in NLTE because the grey cut of cells with tau>8 is very low for the photospheric phase. I suppose this helps speed up the tests so this might not be something we want to change. However in this cell the He fraction is so low that it doesn't get included when setting up the Spencer-fano solution. The simplest change I could make to the test which shouldn't alter the run time of the tests is to just increase the He fraction in the cell that is treated in NLTE in the tests. Just wanted to check if you think this is the best approach or would you recommend something else?
> 
> Can you get something working here, we'll merge it, and then do a separate PR that fixes the test with no code changes? If the checksums need to change, then we don't want to be confused about exactly what caused the change (e.g. unrelated code in the diff).

Yes no problem. This branch is ready for review now if we don't want to make any changes to the tests

lukeshingles (approved):

PR #203, opened 2025-06-23
Update nltephotospheric tests to include Fe VI and VII and He in atomic data

fionntancallan:
Previously none of the automated tests included He or instates higher than V for any of the IGEs. Including He ensures there is a test to ensure we can handle the case when a species has zero bound electrons. Including Fe VI and VII provides a more robust test of the dynamic ion treatment in the photospheric phase when the ionisation state tends to be higher.


tests/nltephotospheric_dynamic_ion_range_1d_1dgrid_inputfiles/abundances.txt
lukeshingles, line 1: Why was this file added?

fionntancallan: It has an increased fraction of He in the cell in this test that is treated in NLTE so that He isn't neglected from the Spencer fans solution due to it having too low of a population

lukeshingles: Ok, makes sense

tests/setup_nltephotospheric_dynamic_ion_range_1d_1dgrid.sh
```diff
@@ -6,21 +6,21 @@ runfolder=nltephotospheric_dynamic_ion_range_1d_1dgrid_testrun
...
-if [ ! -f atomicdata_feconi.tar.xz ]; then curl -O https://theory.gsi.de/~lshingle/artis_http_public/artis/atomicdata_feconi.tar.xz; fi
+if [ ! -f atomicdata_hefeconi_fe_I_to_VII.tar.xz ]; then curl -O https://theory.gsi.de/~lshingle/artis_http_public/artis/atomicdata_hefeconi_fe_i_to_vii.tar.xz; fi
 
-tar -xf atomicdata_feconi.tar.xz --directory $runfolder
+tar -xf atomicdata_hefeconi_fe_i_to_vii.tar.xz --directory $runfolder
 
 rsync -av --ignore-times nltephotospheric_dynamic_ion_range_1d_1dgrid_inputfiles/ $runfolder
 
-rsync -av nebular_1d_3dgrid_inputfiles/abundances.txt nebular_1d_3dgrid_inputfiles/model.txt nebular_1d_3dgrid_inputfiles/recombrates.txt $runfolder
+rsync -av nebular_1d_3dgrid_inputfiles/model.txt $runfolder
```
lukeshingles: It's not new to this PR, but it looks like these lines got swapped around? Normally we'd copy the whole folder and the patch it with updates, like this in setup_nebular_1d_3dgrid_limitbfest.sh:
```sh
rsync -av nebular_1d_3dgrid_inputfiles/ nebular_1d_3dgrid_limitbfest_testrun/
rsync --ignore-times -av nebular_1d_3dgrid_limitbfest_inputfiles/ nebular_1d_3dgrid_limitbfest_testrun/
```
For example, recombrates.txt was not copied and it's not clear whether this is intentional or not.

tests/setup_nltephotospheric_dynamic_ion_range_1d_1dgrid.sh
```diff
@@ -6,21 +6,21 @@ runfolder=nltephotospheric_dynamic_ion_range_1d_1dgrid_testrun
...
 
 cp ../data/* $runfolder
 
 cp ../artisoptions_nltephotospheric_dynamic_ion_range.h $runfolder/artisoptions.h
 
 cd $runfolder
 
-sed -i'' -e 's/constexpr int MPKTS.*/constexpr int MPKTS = 250;/g' artisoptions.h
+sed -i'' -e 's/constexpr int MPKTS.*/constexpr int MPKTS = 20;/g' artisoptions.h
```
lukeshingles: This is far too few packets. If you inspect the logs, you'll see that most timesteps have zero decaying pellets. Increase until the time spent updating packets becomes comparable to the time spent updating the grid.

tests/setup_nltephotospheric_dynamic_ion_range_1d_1dgrid.sh
fionntancallan: Ok that makes sense, I have reordered these lines and made it clear than excluding recombrates.txt is intentional

tests/setup_nltephotospheric_dynamic_ion_range_1d_1dgrid.sh
fionntancallan: When I up the packets to 1000 per process (increase of factor of 50 in packet number) the time to update packets stays negligible (less than 1 second for all cells). However for the second NLTE timestep (the first NLTE timestep where the detailed radiation field and also the detailed BF estimators are turned on) the time to update grid increases by roughly a factor of 25 which makes the test take way too long. The majority of the "time spent" according to the log files is on the NLTE populations (roughly 150s per iteration). Am I missing something here? It isn't clear to me why increasing the packet number would not impact the update packets time but significantly increases the update grid time?

lukeshingles: The grid solution is done on information collected by the packets, so perhaps the NLTE iteration count got very high. You could try manually reducing NLTEITER. Can you point me to the GH Actions log from where you tried this?

fionntancallan: The highest the iteration count ever gets is 7, its the time spent on the population solver call for each element that seems to be increasing the run time significantly. From further testing it is turning on the detailed radiation field specifically which is slowing things down. Is there a reason that would be particularly impacted by the packet number? Currently I am using 256 radiation field bins which may also be high for testing purposes? I have been carrying out these tests locally but can push the test with increased packets to GitHub if you want to look at the results?

fionntancallan: Reducing the number of radfield bins from 256 to 32 very significantly reduces the run time. Is this a reasonable number to use for the test. It's still not clear to me why increasing the packet count leads to such a massive increase in the time taken for update grid when the detailed radfield model is on but maybe thats expected? With 32 radfield bins and maybe setting the max nlte iterations to 3 or 4 I can probably get the test roughly around 10 minutes if those are reasonable choices for those options at least for a test

lukeshingles (approved):

PR #206, opened 2025-06-26
Add ION_NLEVELS_NLTE to artisoptions

fionntancallan:
Add ION_NLEVELS_EXCITED_NLTE function which species the number of NLTE levels for each ion. The return of this function is passed to LEVEL_IS_NLTE which tags the levels that are treated in full NLTE.


artisoptions_christinenonthermal.h
```diff
@@ -16,11 +16,15 @@ constexpr bool FORCE_SPHERICAL_ESCAPE_SURFACE = false;
 
 constexpr int NLTEITER = 30;
 
-constexpr bool LEVEL_IS_NLTE(int element_z, int ionstage, int level) {
+constexpr int ION_NLEVELS_NLTE(int element_z, int ionstage) {
```
lukeshingles: This definition (which includes the ground state) doesn't match get_nlevels_excited_nlte(). I suggest either renaming to ION_NLEVELS_EXCITED_NLTE (and subtracting one from the results), or changing the definition of get_nlevels_excited_nlte() to get_nlevels_nlte() to include ground state (and subtracting one everywhere that the old function was used).

artisoptions_nltephotospheric_dynamic_ion_range.h
```diff
@@ -31,7 +35,7 @@ constexpr bool single_level_top_ion = false;
 
 constexpr bool single_ground_level = false;
 
-constexpr int NLEVELS_REQUIRETRANSITIONS(int Z, int ionstage) { return (Z < 20) ? 100 : 200; }
+constexpr int NLEVELS_REQUIRETRANSITIONS(int Z, int ionstage) { return ION_NLEVELS_NLTE(Z, ionstage) + 1; }
```
lukeshingles: I'm surprised that this didn't affect the checksums since you've increase the values by one on all elements. I suppose that means that either atomic data already had these transitions, or you've limited the max level count in compositiondata.txt to below this number anyway?

lukeshingles (changes requested):
Should be easy to get this merged today, but a couple of small requests.

artisoptions_christinenonthermal.h
fionntancallan: I'm not sure I fully understand the problem here. E.g if the condition for which levels are NLTE was previously:

constexpr bool LEVEL_IS_NLTE(int element_z, int ionstage, int level) {
  if (element_z < 20) {
    return (level <= 100);
  }
  return (level <= 200);
  

Is this not equivalent in terms of what is returned by LEVEL_IS_NLTE:
  
  constexpr int ION_NLEVELS_NLTE(int element_z, int ionstage) {
  if (element_z < 20) {
    return 100;
  }
  return 200;
}

constexpr bool LEVEL_IS_NLTE(int element_z, int ionstage, int level) {
  return level <= ION_NLEVELS_NLTE(element_z, ionstage);
}

artisoptions_nltephotospheric_dynamic_ion_range.h
fionntancallan: I am also surprised by this. I haven't changed anything in the compositiondata.txt to reduce the levels and there are definitely some levels which don't have transitions. within the first 200 levels of the IGEs so not sure how this is possible

fionntancallan: I have now changed this option back to how it was before as NLEVELS_REQUIRETRANSITIONS is set in the setup script for this test so changing it doesn't impact the results of the tests.

lukeshingles (dismissed):

PR #207, opened 2025-06-27
Fix no LTE reset when NLTE_LIMIT_ION_STAGES_AFTER_FAILURE can not find solution

fionntancallan:
There was a bug in the solve_nlte_pops_element function when NLTE_LIMIT_ION_STAGES_AFTER_FAILURE=True that meant when it was not possible to cut either the top or bottom ions the function was returning early before the populations of the element could be set to LTE. Have removed this return statement so the function behaves as intended.


lukeshingles (approved):

lukeshingles (changes requested):
Should the return be replaced with ```matrix_solve_success = false```?

lukeshingles (approved):

PR #208, opened 2025-06-27
Simplify NLTE solver

lukeshingles:


lukeshingles:
@fionntancallan The diff is probably too huge to go line by line here, but could you see if you are happy with the logic in check_can_remove_ion(), and the places in solve_nlte_pops_element() where this function is called? I know that some of these cases won't be part of the test, so some manual inspection is needed.

fionntancallan:
@lukeshingles you mentioned in your comment that I should check the logic of "check_can_remove_ion()" but I'm not seeing that function anywhere. Has that function since been removed or renamed to something else?

lukeshingles:
Please do not make any commits to my branch.

fionntancallan:
> Please do not make any commits to my branch.

Sorry, I just updated it to develop so I could check the printouts in the automated test match the logic I expect - the log files from the previous test had expired. I'm checking through the logic of the refactored code now

lukeshingles:
> > Please do not make any commits to my branch.
> 
> Sorry, I just updated it to develop so I could check the printouts in the automated test match the logic I expect - the log files from the previous test had expired. I'm checking through the logic of the refactored code now

I've rebased it to the latest develop commit now.

nltepop.cc
```diff
@@ -860,105 +959,71 @@ void set_element_pops_lte(const int nonemptymgi, const int element) {
...
+          superlevel_pop, nnelement, NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION);
+      return false;
     }
   }
+
+  printout(
+      "  Passed checks: removing %s ion from NLTE matrix and attempting re-solve (ionstage %d, "
+      "ground pop %g, fabs(excited pops) sum %g, superlevel pop %g, nnelement %g) matrix\n",
+      ionname.c_str(), get_ionstage(element, max_ion_used), ground_pop, nlte_excited_pop_sum, superlevel_pop,
+      nnelement);
```
fionntancallan: I dont think "get_ionstage(element, max_ion_used)" makes sense as the top ion will be printed regardless if it is the top ion being removed. e.g. here is a printout from one of the tests:

Passed checks: removing bottom ion from NLTE matrix and attempting re-solve (ionstage 5, ground pop -1.12046e-09, fabs(excited pops) sum 1.049e-09, superlevel pop -5.53496e-09, nnelement 

 I think there is enough context from surrouding printouts that the best thing to do here is just remove it from the printout unless you have some other reason for it being printed.

nltepop.cc
```diff
@@ -860,105 +959,71 @@ void set_element_pops_lte(const int nonemptymgi, const int element) {
...
+    if (superlevel_pop / nnelement > NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION) {
       printout(
-          "  WARNING: NLTE solver gave negative population to index %zu (Z=%d ionstage %d level %d), pop = %g. "
-          "Replacing with LTE pop of %g\n",
-          row, get_atomicnumber(element), get_ionstage(element, ion), level,
-          gsl_vector_get(&x, row) * pop_normfactors[row], pop_normfactors[row]);
-      popvec[row] = pop_normfactors[row];
+          "  WARNING: Bottom ion superlevel population too large to remove ion (superlevel_pop_bottom_ion / "
+          "nnelement (%g/%g) > (%g) NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION)\n",
+          superlevel_pop, nnelement, NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION);
```
fionntancallan: This says "Bottom ion superlevel" but it should be covering both bottom and top ion cases so I think this needs to be updated to match the printout above and use ionname.c_str(), something like:

      printout(
          "  WARNING: %s ion superlevel population too large to remove ion (superlevel_pop_bottom_ion / "
          "nnelement (%g/%g) > (%g) NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION)\n",
          ionname.c_str(), superlevel_pop, nnelement, NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION);

fionntancallan (commented):
All of the logic seems to match what I intended. I just had a couple of comments on the diagnostic printouts to make sure they are printing out the correct information. At some point it would be great if we could get a test that would trigger a solver fail for the bottom ion so that we have a test case for that as well but I at least was out of ideas on how to achieve that with a test that is reasonable and runs in a fast enough time - so I suppose we will just have to live with that for now. Other than that happy to approve once the couple of small changes to the printouts are added

lukeshingles:
Thanks for the review. I've made your suggested changes.

fionntancallan:
> Thanks for the review. I've made your suggested changes.

Ok great, I will double check the printouts are behaving as intended once the tests finish and then approve the pull request

nltepop.cc
```diff
@@ -1011,19 +1009,13 @@ auto can_remove_ion(const int element, const int ion, const int first_ion_used,
...
+          "  WARNING: %s ion superlevel population too large to remove ion (superlevel_pop_bottom_ion / "
           "nnelement (%g/%g) > (%g) NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION)\n",
-          superlevel_pop, nnelement, NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION);
+          ionname.c_str(), superlevel_pop, nnelement, NLTE_LIMIT_ION_STAGES_MAX_LEVELPOP_OVER_ELEMENTPOP_REMOVE_ION);
       return false;
     }
   }
 
-  printout(
-      "  Passed checks: removing %s ion from NLTE matrix and attempting re-solve (ionstage %d, "
```
fionntancallan: Sorry should have been more precise with my wording here. I meant to remove "get_ionstage(element, max_ion_used)" from this printout. I would still like to keep the rest of the printout

lukeshingles: Ok, how is that now?

fionntancallan: That looks good, will approve now

fionntancallan (approved):

PR #263, opened 2025-11-28
Add compile time option to disable transition data checks

AlexHls:
The old big_gf-4 atomic data set format is not compatible with the current transition data reader and fails the assertion checks. The less restrictive reader in #v2024.08 worked fine. This patch adds a compile flag to skip the more restrictive assertion checks of the latest version and instead uses the ones from #v2024.08.

### DIFF
--- a/artisoptions_classic.h
+++ b/artisoptions_classic.h
@@ -7,6 +7,8 @@
 
 #include "constants.h"
 
+#undef BIG_GF4_LEGACY_COMPAT
+
 constexpr int MPKTS = 100000;
 
 constexpr auto GRID_TYPE = GridType::CARTESIAN3D;
--- a/artisoptions_doc.md
+++ b/artisoptions_doc.md
@@ -1,4 +1,8 @@
 ```
+// Enable/ disable compatibility mode for the old big_gf-4 atomic data set. Removes restrictive checks for transition data
+#undef BIG_GF4_LEGACY_COMPAT
+
+
 // Number of energy packets per process (MPI rank). OpenMP threads share these packets
 constexpr int MPKTS;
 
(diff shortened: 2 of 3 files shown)


lukeshingles:
@AlexHls Thanks for making a PR. I guess the problem is that the transitions is that file are unsorted? The checks were added because the logic when adding extra missing transitions (nlevels_requiretransitions > 0) depends on a the table being sorted by lower, upper ascending. 

I'm trying to minimise the number of artisoptions.h constants as much as possible, since these are becoming a bit unmanageable. I can think of two ways that avoid adding an extra option:

- Read all ion transitions, sort the table, and then apply the nlevels_requiretransitions step.
- Only enforce the lower >= prev_lower and (upper >= prev_upper || lower > prev_lower) checks if nlevels_requiretransitions > 0. After that I would still prefer to sort the table in case we've assumed somewhere else in the code that it is sorted.

The table can be sorted with:
```c++
std::ranges::sort(iontransitiontable, std::less<>{},
                    [](const IonTransitionsInput& t) { return std::tie(t.lower, t.upper); })
```

AlexHls:
I am not quite sure what the problem is. If is actually only a sorting issue, the I would prefer to solve this issue on the atomic data file side, rather than making the file loader more complex. Having more conditionals in the loader seems like a future source of error by introducing more potential edge cases.

The only problem here could be that it seems we can't regenerate the big_gf-4 data set with ``artisatomic``, specifically the public ``gfall`` data. So I guess someone would need to do some reformatting of the existing data files instead. Until then it should be an easy enough mokey-patch to just change these lines. (I assume that I'm the only person using the big_gf-4 data right now.)

lukeshingles:
The crash with this atomic dataset (which contains duplicate transitions) is now fixed on the develop branch. While working on that, I found a bug in the way that we were ensuring that the lowest levels were fully connected to other levels, which will be fixed by #270

PR #275, opened 2026-01-14
Add microclumping

AMG3141:
# Microclumping
Function `clumping_factor` takes time and radial velocity and returns a clumping factor in the interval $[1,\infty)$. These are set in `update_grid`.

## Storing Clumping Factors
New grid property `grid::clumpfactor_allcells`. If `USE_MICROCLUMPING` is set to `false` it is never allocated and a global clumping factor of `1.` is assumed.

## Using Clumping Factors
If microclumping is switched on these processes are affected:
- Collisional (de)excitation
- Collisional ionisation
- Collisional recombination
- Radiative recombination
- Collisional capture
- Stimulated recombination
- Free-free heating
- Free-free cooling

Other processes are physically affected by clumping (e.g. non-thermal processes) but the effects balance eachother out so no changes are required.


lukeshingles:
Still in draft, but mostly for my own curiosity:

- Where are the files that contain clumping factors for every cell and timestep coming from? If these are algorithmically generated, then couldn't the whole design be simplified by defining a new function in artistioptions that takes time_days, radial_velocity, etc and returns the clumping factor? That would avoid the complexity and performance cost of the file I/O, additional configuration options, and allow the same clumping function to be applied to different models and timestep choices.
- Why is "fv" used for the variable names? Is that a volume filling factor? Why not something self-descriptive like "clumpfactor"?
- Is it necessary to pass clumping factors down the call graph or could you simply pass the clumped free electron density to the existing functions, col_excitation_ratecoeff, rad_recombination_ratecoeff, col_recombination_ratecoeff, etc?

Some of the recent updates on the develop branch are showing up in the diff on this PR, which means something weird happened during rebasing. Shouldn't affect merging (since we squash merge anyway) but makes it a bit harder to read the diff.

AMG3141:
It never occurred to me to have a function in artisoptions. I don’t think that would restrict our ability to set the clumping factors in the way Stuart was thinking and it's obviously much more efficient. I will discuss it with him.

Fv is for volume filling factor but was more of a place holder. Something more descriptive is probably the way to go.

I think passing the clumping factor is easier, because it doesn’t require one to think about it each time those functions are used. But I have just noticed that right now I am applying the clumping factor at the return statement(s), but it would require less calls to `grid::apply_clumping` and would probably be easier to read to just have something like `nne = grid::apply_clumping(nne, clumpfactor)` at the start of the functions which require it. So the clumped free electron density is what is used, but doesn't have to be calculated manually each time the function is called.

And yes I messed up rebasing at some point, sorry about that!

lukeshingles:
Ok, sounds good. For now, I recommend implementing it in the simplest way with the minimal code changes to give you the highest confidence that the solution is correct. After that, we can lock the checksums and make sure that the default case of constexpr float clumping_factor() {return 1.0} is being fully optimised out. The rate coefficients are called in tight loops in the macro atom and level pop solver are and fairly critical to performance.

See ION_NLEVELS_EXCITED_NLTE() and NLEVELS_REQUIRETRANSITIONS() for examples of the constexpr functions. You should be able to pass in whatever you need (time_days, radial_velocity) as an argument, but if that's too limiting, you could use a #define that can call anything that is defined in the code. Let me know if you need any help with getting that to work.

AMG3141:
The way I have implemented it currently the clumping factors are read in during `update_grid` and stored as a grid property. `get_oneoverfv` is then a getter for `oneoverfv_allcells`. Is it best to keep this structure (but instead of reading we calculate using the new function) or to calculate the clumping factors as we need them? I guess it’s a trade off between increased memory usage vs more operations?

lukeshingles:
> The way I have implemented it currently the clumping factors are read in during `update_grid` and stored as a grid property. `get_oneoverfv` is then a getter for `oneoverfv_allcells`. Is it best to keep this structure (but instead of reading we calculate using the new function) or to calculate the clumping factors as we need them? I guess it’s a trade off between increased memory usage vs more operations?

I don't think it's obvious which way to go on that, so choose with whichever you find easier to implement. There could be a performance benefit to storing and communicating the clumping factor of each cell if calculating them requires fetching several other numbers (e.g. the cell radial velocity is not stored directly, so has to be calculated from x-y-z coords, in case you want a radial dependence in the clumping factor), but I don't think it's easy to predict which will be faster. One nice thing about calculating it each time is that the compiler can see if your clump factor function always returns 1.00 and will just optimise it out without you having to have to define `bool USE_MICROCLUMPING = false`, whereas it has no idea what numbers will be in the array of clumping factors.

grid.cc
```diff
@@ -384,7 +384,11 @@ void allocate_nonemptymodelcells() {
   kappagrey_allcells = MPI_shared_malloc_span<float>(nonempty_npts_model, 0.);
   grey_depth_allcells = MPI_shared_malloc_span<float>(nonempty_npts_model, 0.);
   thick_allcells = MPI_shared_malloc_span<int>(nonempty_npts_model, 0);
-  const auto modelgrid_mem_usage = nonempty_npts_model * ((sizeof(float) * 9) + sizeof(double) + sizeof(int));
+  if constexpr (USE_MICROCLUMPING) {
+    clumpfactor_allcells = MPI_shared_malloc_span<float>(nonempty_npts_model, -1.);
```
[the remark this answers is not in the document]
AMG3141: When testmode is on get_clumpfactor asserts that the value is in the valid range. This should catch any fundamental problems in the implementation during the testing phase. Will add a check each timestep that ensures none of the clumping factors are negative (though this is something that should probably be verified by the user beforehand).

sn3d.cc
```diff
@@ -711,6 +715,15 @@ auto main(int argc, char* argv[]) -> int {
     vpkt::nvpkt_esc_from_macroatom = 0;
   }
 
+#ifdef READ_CLUMPING_FACTORS_FROM_FILE
+  bool clumping_factors_exists = false;
+  for (auto dir : datafolders) {
+    clumping_factors_exists =
+        clumping_factors_exists || std::filesystem::exists(std::format("{}/clumping-factors.txt", dir));
+  }
+  assert_always(clumping_factors_exists);
```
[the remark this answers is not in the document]
AMG3141: I think it's probably best to have READ_CLUMPING_FACTORS_FROM_FILE as a constexpr bool rather than a macro (even if it does add yet another option...)

sn3d.cc
```diff
@@ -711,6 +726,15 @@ auto main(int argc, char* argv[]) -> int {
...
+#endif
+
```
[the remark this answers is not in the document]
AMG3141: The thinking behind having a dedicated check is so the user sees immediately that there is something wrong rather than having to wait until fstream_required tries to open it in the first update_grid call.

update_grid.cc
```diff
@@ -371,6 +380,19 @@ void update_grid_cell(const int nonemptymgi, const int nts, const int nts_prev,
   const auto rho = static_cast<float>(grid::get_rho_tmin(mgi) / pow(tratmid, 3));
   grid::set_rho(nonemptymgi, rho);
 
+  // Update clumping factors
+  if constexpr (USE_MICROCLUMPING) {
+    float clump_factor;
+#ifdef READ_CLUMPING_FACTORS_FROM_FILE
+    clumping_factors_file >> clump_factor;
+#else
```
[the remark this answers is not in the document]
AMG3141: I was only thinking about MPI when implementing this. Is it worth the effort to write an OpenMP version as well? Would it be acceptable to just say this specific part of the feature (reading clumping factors from a file) is not permitted for OpenMP builds?

update_grid.cc
```diff
@@ -576,6 +598,13 @@ void update_grid(std::ostream& estimators_file, const int nts, const int nts_pre
 
   const auto nstart_nonempty = grid::get_nstart_nonempty(my_rank);
 
+#ifdef READ_CLUMPING_FACTORS_FROM_FILE
+  clumping_factors_file = fstream_required("clumping-factors.txt", std::ios::in);
+  // TODO: this assumes you know beforehand how many nonempty cells there are. Don't know if that's someting that can be
```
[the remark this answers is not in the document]
AMG3141: Will be fixed when changing READ_CLUMPING_FACTORS_FROM_FILE from a macro to a constexpr bool.

update_grid.cc
```diff
@@ -584,6 +627,12 @@ void update_grid(std::ostream& estimators_file, const int nts, const int nts_pre
                      heatingcoolingrates_thisrankcells.at(nonemptymgi - nstart_nonempty));
   }
```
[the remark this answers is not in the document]
AMG3141: I was only thinking about MPI when implementing this. Is it worth the effort to write an OpenMP version as well? Would it be acceptable to just say this specific part of the feature (reading clumping factors from a file) is not permitted for OpenMP builds?

update_grid.cc
```diff
@@ -357,6 +360,13 @@ static void titer_average_estimators(const int nonemptymgi) {
 }
 #endif
 
+// Assumes clumping-factors.txt is formatted according to description in artisoptions_doc.md
+void set_clumping_factors_file_pointer(const int nts, const int mgi) {
+  assert_testmodeonly(nts >= 0 && 0 <= mgi && mgi < grid::get_npts_model());
+  assert_testmodeonly(clumping_factors_file.is_open());
+  clumping_factors_file.seekg(((nts * grid::get_npts_model()) + mgi) * (2 + 6 + 5));
+}
```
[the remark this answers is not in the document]
AMG3141: This works assuming the file is formatted correctly as described (and with Unix line endings). In practice clumping-factors.txt will be generated by a script, so can add checks for correctness in there.

Obviously a binary format would negate the need to worry about any of this, though would be much more difficult to edit manually.

AMG3141:
There are a few things copilot has brought up. I left the unresolved for you to have a look at. A couple of them I've left replies on explaining my thinking behind it. There are also a few things I want to check:
- `grid::get_modelcell_mean_radial_vel`: Any preference on which version to use? I think the commented out one is cleaner, but the one I’ve left in should be slightly more efficient.
- `ratecoeff.cc:652` have I done that correctly? The ternary operator is just there to check for empty vs nonempty cells right?
- `sn3d.cc:560` I’ve added a check that all of the clumping factors are physically sensible values. The thinking is to catch quickly if any clumping factors are problematic, though it does require looping over all non-empty cells. Is it fine to keep it in production or should it be moved to test mode?

One last thing is because copilot automatically reviews, it comments on the same things multiple times and causes a lot of clutter and makes things difficult to follow. Is there a way to disable the automatic reviews?

lukeshingles:
I’m on leave until April 20. If you’re reasonably confident that the implementation is correct then you can update the checksums to avoid breakage as you work on simplifying the design.

grid.h
```diff
@@ -124,6 +129,15 @@ inline auto get_ejecta_kinetic_energy() {
   return E_kin;
 }
 
+// Applies the clumping factor to `val` if microclumping is being used
+template <typename T>
+inline auto apply_clumping(const T val, const float clumpfactor) -> T {
+  if constexpr (USE_MICROCLUMPING) {
+    return val * clumpfactor;
+  }
+  return val;
```
lukeshingles: No need to define a new function just to multiply two numbers together. Remove

grid.h
```diff
@@ -93,6 +96,8 @@ void write_grid_restart_data(int timestep);
 [[nodiscard]] auto get_propcell_random_position_tmin(int cellindex) -> Vec3d;
 [[nodiscard]] DEVICE_FUNC auto boundary_distance(const Vec3d& dir, const Vec3d& pos, double tstart, int cellindex)
     -> std::tuple<double, int>;
+void set_clumpfactor(int nonemptymgi, float vol_filling_factor);
```
lukeshingles: The function is called `set_clumpfactor` but it takes a volume filling factor instead. Rename to 'set_vol_filling_factor' or take the clumping factor as an argument to be consistent.

grid.cc
```diff
@@ -1532,6 +1548,16 @@ auto get_rho_tmin(const int modelgridindex) -> float { return modelgrid_input[mo
...
+[[gnu::pure]] [[nodiscard]] DEVICE_FUNC auto get_modelcell_mean_radial_vel(const int modelgridindex, const double tmin)
+    -> double {
+  // TODO: both versions give the same thing. uncommented one should give ever so slightly better performance
+
+  const int assoc_cells = grid::get_numpropcells(modelgridindex);
+  return modelgrid_input[modelgridindex].initial_radial_pos_sum / tmin / assoc_cells;
+  // ---------------
+  // return get_modelcell_mean_radial_pos(modelgridindex, tratmid) / tmid;
+}
+
```
lukeshingles: Trivial function used only once. Inline at the call site.

grid.cc
```diff
@@ -1688,6 +1720,18 @@ DEVICE_FUNC auto get_modelgridtype() -> GridType {
...
+// `nonemptymgi` is set to -1
+[[nodiscard]] DEVICE_FUNC auto check_mgi_is_nonempty(const int mgi, int& nonemptymgi) -> bool {
+  assert_testmodeonly(get_nonempty_npts_model() > 0);
+  assert_testmodeonly(mgi >= 0);
+  assert_testmodeonly(mgi < get_npts_model());
+
+  nonemptymgi = nonemptymgi_of_mgi[mgi];
+  return nonemptymgi >= 0;
+}
+
```
lukeshingles, line 1714: Other parts of the code check if a cell is non-empty using the expression: `grid::get_numpropcells(mgi) > 0`

artisoptions_christinenonthermal.h
```diff
@@ -159,5 +159,11 @@ constexpr auto GAMMA_THERMALISATION_SCHEME = GammaThermalisationScheme::FREQUENC
 
 constexpr std::optional<double> GAMMA_USE_KAPPA_GREY;
 
+constexpr bool USE_MICROCLUMPING = false;
+
+constexpr bool READ_VOLUME_FILLING_FACTORS_FROM_FILE = false;
```
lukeshingles: I thought we had eliminated the need for file I/O using the artisoptions function? Can that be removed now?

kpkt.cc
```diff
@@ -52,6 +52,7 @@ auto calculate_cooling_rates_ion(const int nonemptymgi, const int element, const
                                  double* const C_exc, double* const C_ionisation) -> double {
   const auto nne = grid::get_nne(nonemptymgi);
   const auto T_e = grid::get_Te(nonemptymgi);
+  const float clumpfactor = grid::get_clumpfactor(nonemptymgi);
```
lukeshingles: Replace with:
const auto clumpednne = nne * grid::get_clumpfactor(nonemptymgi);
Then use clumpednne instead of nne everywhere below.

kpkt.cc
```diff
@@ -63,7 +64,7 @@ auto calculate_cooling_rates_ion(const int nonemptymgi, const int element, const
   const int ioncharge = get_ionstage(element, ion) - 1;
   if (ioncharge > 0) {
     const double C_ff_ion = 1.426e-27 * sqrt(T_e) * pow2(ioncharge) * nncurrention * nne;
```
lukeshingles: Does nncurrention also need a clumping factor?

macroatom.h
```diff
@@ -16,21 +17,24 @@ DEVICE_FUNC void do_macroatom(Packet& pkt, const MacroAtomState& pktmastate);
...
 
 [[gnu::pure]] [[nodiscard]] auto col_deexcitation_ratecoeff(float T_e, float nne, double epsilon_trans,
                                                             double upperstatweight, double lowerstatweight,
-                                                            int alltransindex) -> double;
+                                                            int alltransindex, float clumpfactor) -> double;
 
 [[gnu::pure]] [[nodiscard]] auto col_excitation_ratecoeff(float T_e, float nne, double upperstatweight,
                                                           int alltransindex, double epsilon_trans,
-                                                          double lowerstatweight) -> double;
+                                                          double lowerstatweight, float clumpfactor) -> double;
```
lukeshingles: No need to change these function signatures - just pass in the local electron density with clumping applied. Maybe rename 'nne' parameters to 'clumpednne' or 'localnne'

kpkt.cc
AMG3141: It's contained within nne. Renamed variable to clumpednne to make that clearer.

ratecoeff.cc
```diff
@@ -426,15 +426,16 @@
   // stimulated recombination is negative photoionisation
   const double nnlevel = use_cellcache ? get_cellcache_levelpop(nonemptymgi, loweruniquelevelindex)
                                        : calculate_levelpop(nonemptymgi, element, ion, level);
-  const double nne = grid::get_nne(nonemptymgi);
+  const double clumpednne = grid::get_nne(nonemptymgi) * grid::get_clumpfactor(nonemptymgi);
```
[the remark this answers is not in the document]
AMG3141: @lukeshingles This call to get_nne was originally a double so I left it as such. I'm not really sure how big nne tends to be. For nominal values of nne would there be any risk of overflowing float when applying a clumping factor? If so we could change all the clumpednnes to be doubles. Otherwise just change this to float for consistency.

artisoptions_christinenonthermal.h
AMG3141: I'll remove it for now and if it ends up being something that might be useful I can put it back

ratecoeff.cc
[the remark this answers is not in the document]
lukeshingles: Overflow is very unlikely here, but to avoid the warnings, you can change it to:
```c++
const auto clumpednne = grid::get_nne(nonemptymgi) * grid::get_clumpfactor(nonemptymgi);


```

sn3d.cc
```diff
@@ -652,6 +656,17 @@ auto do_timestep(const int nts, const int titer, std::vector<Packet>& packets, c
...
+  if constexpr (USE_MICROCLUMPING) {
+    const int nstart_nonempty = grid::get_nstart_nonempty(globals::my_rank);
+    const int ndo_nonempty = grid::get_ndo_nonempty(globals::my_rank);
+
+    for (int nonemptymgi = nstart_nonempty; nonemptymgi < (nstart_nonempty + ndo_nonempty); nonemptymgi++) {
+      const float clump_factor = grid::get_clumpfactor(nonemptymgi);
+      assert_always(std::isfinite(clump_factor) && clump_factor >= 1.F);
+    }
+  }
+
```
lukeshingles: Move this check to set_clumpfactor()

ratecoeff.cc
```diff
@@ -616,6 +618,9 @@ auto calculate_ionrecombcoeff(const int nonemptymgi, const float T_e, const int
 
   // this gets divided and cancelled out in the radiative case anyway
   const auto nne = (nonemptymgi >= 0) ? grid::get_nne(nonemptymgi) : 1.F;
+
+  // TODO: is this the correct way to do this? (Just checking that the cell is non-empty I think)
```
lukeshingles: Yes, looks good. Remove this todo comment.

macroatom.cc
```diff
@@ -638,7 +640,7 @@ void macroatom_open_file() {
 
 // collisional ionisation rate: paperII 3.5.1
 // multiply by lower level population to get a rate per second
-[[gnu::pure]] [[nodiscard]] auto col_ionisation_ratecoeff(const float T_e, const float nne, const int element,
+[[gnu::pure]] [[nodiscard]] auto col_ionisation_ratecoeff(const float T_e, float clumpednne, const int element,
```
lukeshingles: Why did you remove const?

macroatom.cc
```diff
@@ -652,15 +654,16 @@ void macroatom_open_file() {
...
+      clumpednne * 1.55e13 * pow(T_e, -0.5) * g * sigma_bf * exp(-fac1) / fac1;  // photoionisation at the edge
 
   assert_testmodeonly(std::isfinite(C));
 
   return C;
 }
 
 // multiply by upper level population to get a rate per second
-[[gnu::pure]] [[nodiscard]] auto col_deexcitation_ratecoeff(const float T_e, const float nne,
+[[gnu::pure]] [[nodiscard]] auto col_deexcitation_ratecoeff(const float T_e, float clumpednne,
```
lukeshingles: As above

macroatom.cc
```diff
@@ -692,23 +695,23 @@ void macroatom_open_file() {
...
   // positive coll_str is treated as effective collision strength
 
   // from Osterbrock and Ferland, p51
-  return nne * 8.629e-6 * static_cast<double>(coll_strength) / upperstatweight / std::sqrt(T_e);
+  return clumpednne * 8.629e-6 * static_cast<double>(coll_strength) / upperstatweight / std::sqrt(T_e);
 }
 
 // multiply by lower level population to get a rate per second
-[[gnu::pure]] [[nodiscard]] auto col_excitation_ratecoeff(const float T_e, const float nne,
+[[gnu::pure]] [[nodiscard]] auto col_excitation_ratecoeff(const float T_e, float clumpednne,
```
lukeshingles: As above

update_grid.cc
```diff
@@ -5,7 +5,7 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdlib>
-#include <ostream>
+#include <iostream>
```
lukeshingles: Why was this change necessary?

update_grid.cc
```diff
@@ -378,6 +378,15 @@ void update_grid_cell(const int nonemptymgi, const int nts, const int nts_prev,
   const auto rho = static_cast<float>(grid::get_rho_tmin(mgi) / pow3(tratmid));
   grid::set_rho(nonemptymgi, rho);
 
+  // Update clumping factors
+  if constexpr (USE_MICROCLUMPING) {
+    const double tmid = globals::timesteps[nts].mid;
+    const double rad_vel = grid::get_modelcell_mean_radial_pos(grid::get_mgi_of_nonemptymgi(nonemptymgi), tmid) / tmid;
+    const float vol_filling_factor = volume_filling_factor(tmid, rad_vel);
+
+    grid::set_clumpfactor(nonemptymgi, 1.F / vol_filling_factor);
```
lukeshingles: Why is your configuration option "volume_filling_factor" and then you store the reciprocal clump_factor? Why not specify the clump factor directly in the configuration?

macroatom.cc
AMG3141: Left over from when I was passing clumping factor separately and modifying nne

update_grid.cc
AMG3141: Was used when reading from file, not needed anymore

update_grid.cc
AMG3141: Again left over from reading from file

lukeshingles (approved):
Looks good, let's merge.

For later follow-up (should only affect classic mode simulations, not full NLTE): check and update calls to find_uppermost_ion(), find_converged_nne(), and calculate_ionfractions().

PR #288, opened 2026-02-13
Default to enabling NLTE ionisation stage limiter and simplify related options

lukeshingles:


lukeshingles:
Will do this piece-wise later on.

PR #303, opened 2026-02-23
Change integration variable for alpha_sp to avoid non-finite results

AMG3141:
## Theory
Currently $\alpha_i^{\text{sp}}$ is calculated using the Saha factor as

$$\alpha_i^{\text{sp}}=4\pi\Phi_{i\kappa}\int_{\nu_i}^\infty\frac{a_{i\kappa}(\nu)}{h\nu}\frac{2h\nu^3}{c^2}\exp\left(-\frac{h\nu}{k_BT}\right)\text d\nu,$$

where

$$\Phi_{i\kappa}=\frac{g_i}{g_\kappa}C_IT^{-\frac32}\exp\left(\frac{h\nu_i}{k_BT}\right).$$ 

The exponentials can over/under flow when pre-calculating the rate coefficient integrals.

Re-arranging gives

$$\alpha_i^{\text{sp}}=4\pi\Phi_{i\kappa}^*\int_{\nu_i}^\infty\frac{a_{i\kappa}(\nu)}{h\nu}\frac{2h\nu^3}{c^2}\exp\left(\frac{h(\nu_i-\nu)}{k_BT}\right)\text d\nu,$$

where $\Phi_{i\kappa}^*=\frac{g_i}{g_\kappa}C_IT^{-\frac32}.$

We can do a substitution by setting $u=\nu_i-\nu$ then $\text du=-\text d\nu$, $\nu_i\to0$, $\infty\to-\infty$, and $\nu\to\nu_i-\nu$, thus

$$
\begin{align\*}
	\alpha_i^{\text{sp}}&=4\pi\Phi_{i\kappa}^*\int_0^{-\infty}\frac{a_{i\kappa}(\nu_i-u)}{h(\nu_i-u)}\frac{2h(\nu_i-u)^3}{c^2}\exp\left(\frac{hu}{k_BT}\right)\cdot-\text du\\
	&=4\pi\Phi_{i\kappa}^\*\int_{-\infty}^0a_{i\kappa}(\nu_i-u)\frac{2(\nu_i-u)^2}{c^2}\exp\left(\frac{hu}{k_BT}\right)\text du.
\end{align\*}
$$

## Implementation
In `ratecoeff.cc` I changed the calling of the integrator from
```cpp
auto alpha_sp = integrator<alpha_sp_integrand>(intparas, nu_threshold, nu_max_phixs, RATECOEFF_INTEGRAL_ACCURACY, &error);
```
to
```cpp
auto alpha_sp = integrator<alpha_sp_integrand>(intparas, nu_threshold - nu_max_phixs, 0, RATECOEFF_INTEGRAL_ACCURACY, &error);
```
and the integrand from
```cpp
const auto sigma_bf = photoionisation_crosssection_fromtable(photoion_xs, nu_edge, nu);
const double x = TWOOVERCLIGHTSQUARED * sigma_bf * pow(nu, 2) * exp(-HOVERKB * nu / T);
```
to
```cpp
const auto sigma_bf = photoionisation_crosssection_fromtable(photoion_xs, nu_edge, nu_edge - nu);
const double x = TWOOVERCLIGHTSQUARED * sigma_bf * pow(nu_edge - nu, 2) * exp(-HOVERKB * (nu_edge - nu) / T);
```

## Issues
Failing on a few of the tests. Most of them are failing on the checksums, with just classicmode_1d_3dgrid failing on an assertion. Not quite sure what the cause of that is.

Once this is (hopefully) working a similar change can probably be made in the other places where the Saha factor is used.

@lukeshingles if you could take a look at this and tell me what I've done wrong that would be  greatly appreciated 😄


lukeshingles:
The first thing I recommend with a numerical change like this is to log the values of the old and new method side by side to make sure you’re getting similar numbers (except for cases where the old method was non-finite).

ltepop.h
```diff
@@ -32,6 +32,15 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
...
 
+// Calculate Saha factor divided exp(E_threshold / k_B * T)
+[[gnu::const]] [[nodiscard]] constexpr auto calculate_sahafact_over_e_to_hnuoverkt(const double g_lower,
+                                                                                   const double g_upper, const double T)
+    -> double {
+  const double result = SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
+  assert_testmodeonly(std::isfinite(result));
+  return result;
+}
+
```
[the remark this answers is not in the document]
lukeshingles: Probably better to just eliminate this tiny function and just place the expression inline where it is used.

ratecoeff.cc
```diff
@@ -386,18 +386,19 @@ void precalculate_rate_coefficient_integrals() {
             const auto T_e = static_cast<float>(MINTEMP * exp(iter * T_step_log));
 
             const double sahafact = calculate_sahafact(statw_lower, statw_upper, T_e, E_threshold);
+            const double sahafact_modified = calculate_sahafact_over_e_to_hnuoverkt(statw_lower, statw_upper, T_e);
             assert_always(sahafact >= 0.);
-            assert_always(std::isfinite(sahafact));
+            //assert_always(std::isfinite(sahafact));
```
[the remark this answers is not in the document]
AMG3141: Just commented out for testing purposes. Will ensure proper checks once the fix is working.

AMG3141:
After alpha_sp is calculated I logged atomic_number, ionstage, level, phixstargetindex, iter, alpha_sp. Here are the [final combined and sorted output files](https://qubstudentcloud-my.sharepoint.com/:f:/g/personal/40331764_ads_qub_ac_uk/IgDE5JKuEDTaSLJNoqH2286uAXJPpJvu_IGIgV_U3eJackY?e=ovnV90) for the old version (alpha_sp_develop.txt) and the new version (alpha_sp_update-alpha-sp-calculation.txt).

Doing a diff they seem to generally agree to around 3 significant figures and there's no nan/infs in the new version. alpha_sps.tar.gz is just the same thing but compressed for if you want to download them and take a look.

lukeshingles:
> After alpha_sp is calculated I logged atomic_number, ionstage, level, phixstargetindex, iter, alpha_sp. Here are the [final combined and sorted output files](https://qubstudentcloud-my.sharepoint.com/:f:/g/personal/40331764_ads_qub_ac_uk/IgDE5JKuEDTaSLJNoqH2286uAXJPpJvu_IGIgV_U3eJackY?e=ovnV90) for the old version (alpha_sp_develop.txt) and the new version (alpha_sp_update-alpha-sp-calculation.txt).
> 
> Doing a diff they seem to generally agree to around 3 significant figures and there's no nan/infs in the new version. alpha_sps.tar.gz is just the same thing but compressed for if you want to download them and take a look.

Ok, this sounds promising. Just clean up the diff and we can merge soon. Why you are getting non-finite results when there is already an assertion to check this (and it passes everywhere I have seen so far). Is this problem particular to whatever atomic dataset you’re using?

AMG3141:
> > After alpha_sp is calculated I logged atomic_number, ionstage, level, phixstargetindex, iter, alpha_sp. Here are the [final combined and sorted output files](https://qubstudentcloud-my.sharepoint.com/:f:/g/personal/40331764_ads_qub_ac_uk/IgDE5JKuEDTaSLJNoqH2286uAXJPpJvu_IGIgV_U3eJackY?e=ovnV90) for the old version (alpha_sp_develop.txt) and the new version (alpha_sp_update-alpha-sp-calculation.txt).
> > Doing a diff they seem to generally agree to around 3 significant figures and there's no nan/infs in the new version. alpha_sps.tar.gz is just the same thing but compressed for if you want to download them and take a look.
> 
> Ok, this sounds promising. Just clean up the diff and we can merge soon. Why you are getting non-finite results when there is already an assertion to check this (and it passes everywhere I have seen so far). Is this problem particular to whatever atomic dataset you’re using?

I am using the new dataset I've compiled from the most recent cmfgen compilation and I removed the assert so all the coefficients would be calculated. I ran classicmode_1d_3dgrid as well and it agrees to the same levels of precision, but also wanted to check with a larger dataset.

ltepop.h
```diff
@@ -24,6 +24,7 @@ void calculate_cellpartfuncts(int nonemptymgi, int element);
     -> std::vector<double>;
 void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_saha);
 
+// TODO: remove these functions once implementation is done
```
lukeshingles: calculate_sahafact() is called in other places, so we'll keep that. Remove this comment.

ltepop.h
```diff
@@ -32,6 +33,15 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
...
 
+// Calculate Saha factor divided exp(E_threshold / k_B * T)
+[[gnu::const]] [[nodiscard]] constexpr auto calculate_sahafact_over_e_to_hnuoverkt(const double g_lower,
+                                                                                   const double g_upper, const double T)
+    -> double {
+  const double result = SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
+  assert_testmodeonly(std::isfinite(result));
+  return result;
+}
+
```
lukeshingles: Remove

ratecoeff.cc
```diff
@@ -269,10 +269,10 @@ auto alpha_sp_integrand(const double nu, void* const voidparas) -> double {
...
   const auto& photoion_xs = params.photoion_xs;
 
-  const auto sigma_bf = photoionisation_crosssection_fromtable(photoion_xs, nu_edge, nu);
-  const double x = TWOOVERCLIGHTSQUARED * sigma_bf * pow(nu, 2) * exp(-HOVERKB * nu / T);
-  // in formula this looks like
-  // x = sigma_bf/H/nu * 2*H*pow(nu,3)/pow(CLIGHT,2) * exp(-H*nu/KB/T);
+  const auto sigma_bf = photoionisation_crosssection_fromtable(photoion_xs, nu_edge, nu_edge - nu);
+  const double x = TWOOVERCLIGHTSQUARED * sigma_bf * pow(nu_edge - nu, 2) * exp(HOVERKB * nu / T);
+  // with the substitution u = nu_edge - nu (integration variable 'nu' here is u), the formula looks like
+  // x = sigma_bf/H/(nu_edge - nu) * 2*H*pow(nu_edge - nu,3)/pow(CLIGHT,2) * exp(-H*(nu_edge - nu)/KB/T);
```
lukeshingles: I don't think the x = ... line matches the current function. Keep the comment about change of variables, but drop the x= line.

ratecoeff.cc
```diff
@@ -386,18 +386,21 @@ void precalculate_rate_coefficient_integrals() {
             const auto T_e = static_cast<float>(MINTEMP * exp(iter * T_step_log));
 
             const double sahafact = calculate_sahafact(statw_lower, statw_upper, T_e, E_threshold);
+            // const double sahafact_modified = calculate_sahafact_over_e_to_hnuoverkt(statw_lower, statw_upper, T_e);
```
lukeshingles: Delete

ratecoeff.cc
```diff
@@ -386,18 +386,21 @@ void precalculate_rate_coefficient_integrals() {
...
+            const double sahafact_modified = SAHACONST * statw_lower / statw_upper * std::pow(T_e, -1.5);
+            // TODO: asserts and that
             assert_always(sahafact >= 0.);
-            assert_always(std::isfinite(sahafact));
+            // assert_always(std::isfinite(sahafact));
```
lukeshingles: Re-enable this

ratecoeff.cc
```diff
@@ -386,18 +386,21 @@ void precalculate_rate_coefficient_integrals() {
             const auto T_e = static_cast<float>(MINTEMP * exp(iter * T_step_log));
 
             const double sahafact = calculate_sahafact(statw_lower, statw_upper, T_e, E_threshold);
+            // const double sahafact_modified = calculate_sahafact_over_e_to_hnuoverkt(statw_lower, statw_upper, T_e);
+            const double sahafact_modified = SAHACONST * statw_lower / statw_upper * std::pow(T_e, -1.5);
+            // TODO: asserts and that
```
lukeshingles: Remove

ratecoeff.cc
```diff
@@ -386,18 +386,21 @@ void precalculate_rate_coefficient_integrals() {
             const auto T_e = static_cast<float>(MINTEMP * exp(iter * T_step_log));
 
             const double sahafact = calculate_sahafact(statw_lower, statw_upper, T_e, E_threshold);
+            // const double sahafact_modified = calculate_sahafact_over_e_to_hnuoverkt(statw_lower, statw_upper, T_e);
+            const double sahafact_modified = SAHACONST * statw_lower / statw_upper * std::pow(T_e, -1.5);
```
lukeshingles, line 389: It's not really a Saha factor anymore. I suggest just taking this small expression (SAHACONST * statw_lower / statw_upper * std::pow(T_e, -1.5)) and placing it inline at the alpha_sp *=... line. That avoids the problem of naming it and saves one line of code.

ratecoeff.cc
```diff
@@ -386,18 +386,21 @@ void precalculate_rate_coefficient_integrals() {
...
             // the threshold of the first target gives nu of the first phixstable point
             const GSLIntegrationParas intparas = {
                 .nu_edge = nu_threshold, .T_e = T_e, .photoion_xs = get_phixs_table(element, ion, level)};
 
             // Spontaneous recombination and bf-cooling coefficient don't depend on the radiation field
-            auto alpha_sp = integrator<alpha_sp_integrand>(intparas, nu_threshold, nu_max_phixs,
+            auto alpha_sp = integrator<alpha_sp_integrand>(intparas, nu_threshold - nu_max_phixs, 0,
                                                            RATECOEFF_INTEGRAL_ACCURACY, &error);
-            alpha_sp *= FOURPI * sahafact * phixstargetprobability;
+            alpha_sp *= FOURPI * sahafact_modified * phixstargetprobability;
```
lukeshingles: Since we're changing the checksums anyway, can you merge these two lines and make alpha_sp const?

lukeshingles (changes requested):
A small few changes to do and then the next thing is to update the checksums. Go to the output page of the latest CI run and download checksums.zip. Save that to artis/tests, unzip the file, and then commit the changed files.

ratecoeff.cc
```diff
@@ -386,24 +386,26 @@ void precalculate_rate_coefficient_integrals() {
...
 
             if (!std::isfinite(alpha_sp) || alpha_sp < 0) {
               printlnlog(
-                  "WARNING: alpha_sp was negative or non-finite for level {} Te {:g}. alpha_sp {:g} sfac {:g} "
+                  "WARNING: alpha_sp was negative or non-finite for level {} Te {:g}. alpha_sp {:g} sfac_mod {:g} "
                   "phixstargetindex {} phixstargetprobability {:g}",
-                  level, T_e, alpha_sp, sahafact, phixstargetindex, phixstargetprobability);
+                  level, T_e, alpha_sp, sahafact_modified, phixstargetindex, phixstargetprobability);
```
lukeshingles: Can remove sahafact_modified from the log message.

ratecoeff.cc
AMG3141: I think this quantity will have to be used again in this_bfcooling_coeff when that is updated in a similar way? Should it just be inlined there as well?

ratecoeff.cc
AMG3141: It depends if we want to keep the check below and set alpha_sp to 0 if it is non-finite or negative. In my big run with the new cmfgen data this check was never triggered.

ratecoeff.cc
lukeshingles: If you'll be using it multiple times then keep the variable.

ratecoeff.cc
lukeshingles: If you're keeping the variable, then no change is needed here.

ratecoeff.cc
lukeshingles: I think we should just crash if the alpha_sp is invalid.

ratecoeff.cc
lukeshingles: Sorry, I should have noticed this on the previous review. The nu parameter here is not nu now but should be renamed to nu_edge_minus_nu.

It also occured to me that putting in a minus sign so that u = nu - nu_edge might be a little more readable (e.g. the integration limits), but that might make the checksums change again. Something to consider for the next PR if you'll be doing similar transformations.

lukeshingles (changes requested):
One last thing and we can merge. Nice work!

lukeshingles (approved):

PR #306, opened 2026-02-25
Fix (non-default) XCOM gamma photoelectric cross section interpolation

gleck97:
Fix critical small bug in XCOM cross section interpolation.


lukeshingles:
This is clearly not in a state where you should be requesting my review.

lukeshingles:
Why did you close the PR? Just fix the title, run clang-format, update the checksums and we can merge. Request my review when this is done.

lukeshingles (changes requested):
Needs to be updated to the latest develop commit and then it's ready to merge.

lukeshingles (approved):

PR #307, opened 2026-02-25
Change integration variable for bfcooling_coeff to avoid non-finite results

AMG3141:
Similar change for bfcooling_coeff as #303


ratecoeff.cc
```diff
@@ -331,16 +331,17 @@ auto approx_bfheating_integrand(const double nu, void* const voidparas) -> doubl
...
 // the resulting expression with the correct W later on.
-auto bfcooling_integrand(const double nu, void* const voidparas) -> double {
+auto bfcooling_integrand(const double nu_edge_minus_nu, void* const voidparas) -> double {
   const auto& params = *(static_cast<const GSLIntegrationParas*>(voidparas));
   const auto& nu_edge = params.nu_edge;
   const auto& T = params.T_e;
   const auto& photoion_xs = params.photoion_xs;
 
-  const float sigma_bf = photoionisation_crosssection_fromtable(photoion_xs, nu_edge, nu);
+  const float sigma_bf = photoionisation_crosssection_fromtable(photoion_xs, nu_edge, nu_edge - nu_edge_minus_nu);
```
lukeshingles: Can you change the variable to nu_minus_nu_edge instead? This will make the integration limits positive. Might as well fix the alpha_sp one at the same time.

ratecoeff.cc
```diff
@@ -331,16 +331,17 @@ auto approx_bfheating_integrand(const double nu, void* const voidparas) -> doubl
 // on a temperature grid using the assumption that T_e=T_R and W=1 in the ionisation
 // formula. The radiation fields dependence on W is taken into account by multiplying
 // the resulting expression with the correct W later on.
```
AMG3141: This comment seems to be copied exactly from the above function. I assume it should say "bound-free cooling" at the start? Does the rest of the comment apply to this function?

lukeshingles: Good catch. It should say "bound-free cooling" and the radiation field is not relevant here. I'll quickly fix this myself because I want to get your PR merged before I merge a couple of other things that alter checksums.

lukeshingles (dismissed):

lukeshingles (approved):

PR #314, opened 2026-02-27
Update departure ratio calculations to avoid non-finite results

AMG3141:
Similar changes to #303 and #307 for the departure ratios.

## Corrphotoioncoeff
Using the same substitution ($\nu-\nu_i$) as the previous two changes we can write

$$\int_{\nu_i}^\infty\frac1h\frac\sigma\nu J_\nu\left(1-d\exp\left(-\frac{h\nu}{k_BT_e}\right)\right)\text d\nu=\int_0^{\infty}\frac1h\frac\sigma{(u+\nu_i)}J_{u+\nu_i}\left(1-d^\prime\exp\left(-\frac{hu}{k_BT_e}\right)\right)\text du,$$

where $d$ ($d^\prime$) is the (modified) departure ratio ($\Phi^\prime$ is the modified Saha factor):

$$
\begin{align*}
d=\frac{n_\kappa n_e}{n_i}\Phi &, & d^\prime=\frac{n_\kappa n_e}{n_i}\Phi^\prime.
\end{align*}
$$

## Chi_bf_gammacontr
Just a rearrangement of the stimfactor calculation:

$$d\exp\left(-\frac{h\nu}{k_BT_e}\right)=d^\prime\exp\left(-\frac{h(\nu-\nu_i)}{k_BT_e}\right)$$


ltepop.h
```diff
@@ -32,6 +32,17 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
   return sahafact;
 }
 
+// calculates saha factor in LTE multiplied by exp(-E_threshold / KB / T): Phi_level,ion,element =
+// nn_level,ion,element/(nne*nn_upper,ion+1,element)
+// avoids possible overflows when using the regular sahafact
+[[gnu::const]] [[nodiscard]] constexpr auto calculate_modified_sahafact(const double g_lower, const double g_upper,
```
lukeshingles: As in other cases, this expression is too simple to be worth a separate function.

ratecoeff.cc
```diff
@@ -440,33 +439,34 @@ auto calculate_corrphotoioncoeff_integral(const int element, const int ion, cons
...
   };
 
   double error = 0.;
 
-  auto gammacorr =
-      integrator<integrand_corrphotoioncoeff_custom_radfield>(intparas, nu_threshold, nu_max_phixs, epsrel, &error);
+  const auto gammacorr = FOURPI * integrator<integrand_corrphotoioncoeff_custom_radfield>(
+                                      intparas, 0, nu_max_phixs - nu_threshold, epsrel, &error);
+
+  // TODO: Just assert here? Since if it goes infinite now something else has probably gone wrong
```
lukeshingles: Yes, assert is better. Silently setting it to zero was a pretty bad idea on my part.

rpkt.cc
```diff
@@ -822,21 +822,24 @@ auto calculate_chi_bf_gammacontr(const int nonemptymgi, const double nu, Phixsli
             photoionisation_crosssection_fromtable(get_phixs_table(allcont_uniquelevelindex[i]), nu_edge, nu);
 
         double corrfactor = 1.;  // default to no subtraction of stimulated recombination
-        double departure_ratio = globals::cellcache[cellcacheslotid].allcont_departureratios[i];
-        if (!USECELLHISTANDUPDATEPHIXSLIST || departure_ratio < 0) {
+        double modified_departure_ratio = globals::cellcache[cellcacheslotid].allcont_modified_departureratios[i];
+
+        // TODO: I don't think this condition needs updating since the only difference between departure_ratio and
```
lukeshingles: Yes, -1 is used as a special value set in cellcache_change_cell() to indicate that the true value has not been calculated yet.

lukeshingles:
> `calculate_corrphotoioncoeff_integral` is called in two places (`calculate_iongamma_per_ionpop` and `get_corrphotoioncoeff`) so I logged them separately ([outputs here](https://qubstudentcloud-my.sharepoint.com/:f:/g/personal/40331764_ads_qub_ac_uk/IgAOebzO_W9DR52BAPoNJetJAUgjXYm_g5iLjDxMVR35IFU?e=PU8ZXV), the files ending in develop are the current implementation, those ending in update-departure-ratio-calculations are the new implementation). I can't see any major differences between the old way and new way.

I see differences of a factor of three of four between the new and old numbers in your files (please attach files directly here instead of linking to OneDrive). Has this been fixed?

ratecoeff.cc
```diff
@@ -440,32 +440,30 @@ auto calculate_corrphotoioncoeff_integral(const int element, const int ion, cons
...
-      integrator<integrand_corrphotoioncoeff_custom_radfield>(intparas, nu_threshold, nu_max_phixs, epsrel, &error);
-  if (!std::isfinite(gammacorr)) {
-    return 0.;
-  }
-
-  gammacorr *= FOURPI * get_phixsprobability(loweruniquelevelindex, phixstargetindex);
+  const auto gammacorr =
+      FOURPI * get_phixsprobability(loweruniquelevelindex, phixstargetindex) *
+      integrator<integrand_corrphotoioncoeff_custom_radfield>(intparas, 0, nu_max_phixs - nu_threshold, epsrel, &error);
+  assert_always(std::isfinite(gammacorr));
```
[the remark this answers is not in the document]
lukeshingles: That's a good catch by copilot. I'll commit a fix for this in your branch.

ratecoeff.cc
```diff
@@ -440,32 +440,30 @@ auto calculate_corrphotoioncoeff_integral(const int element, const int ion, cons
   const double nne = grid::get_nne(nonemptymgi);
   const int upperionlevel = get_phixsupperlevel(loweruniquelevelindex, phixstargetindex);
   const auto upperuniquelevelindex = get_uniquelevelindex(element, ion + 1, upperionlevel);
-  const double sf =
-      calculate_sahafact(stat_weight(loweruniquelevelindex), stat_weight(upperuniquelevelindex), T_e, H * nu_threshold);
+  const double modified_sahafact =
+      SAHACONST * stat_weight(loweruniquelevelindex) / stat_weight(upperuniquelevelindex) * std::pow(T_e, -1.5);
```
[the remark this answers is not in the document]
lukeshingles: I disagree with this suggestion 😊. Resolving

AMG3141:
> > `calculate_corrphotoioncoeff_integral` is called in two places (`calculate_iongamma_per_ionpop` and `get_corrphotoioncoeff`) so I logged them separately ([outputs here](https://qubstudentcloud-my.sharepoint.com/:f:/g/personal/40331764_ads_qub_ac_uk/IgAOebzO_W9DR52BAPoNJetJAUgjXYm_g5iLjDxMVR35IFU?e=PU8ZXV), the files ending in develop are the current implementation, those ending in update-departure-ratio-calculations are the new implementation). I can't see any major differences between the old way and new way.
> 
> I see differences of a factor of three of four between the new and old numbers in your files (please attach files directly here instead of linking to OneDrive). Has this been fixed?

Sorry those are the outputs from last week. Here are updated versions
[calculate_iongamma_per_ionpop_develop.txt](https://github.com/user-attachments/files/25713249/calculate_iongamma_per_ionpop_develop.txt)
[get_corrphotoioncoeff_develop.txt](https://github.com/user-attachments/files/25713251/get_corrphotoioncoeff_develop.txt)
[calculate_iongamma_per_ionpop_update-departure-ratio-calculations.txt](https://github.com/user-attachments/files/25713270/calculate_iongamma_per_ionpop_update-departure-ratio-calculations.txt)
[get_corrphotoioncoeff_update-departure-ratio-calculations.txt](https://github.com/user-attachments/files/25713281/get_corrphotoioncoeff_update-departure-ratio-calculations.txt)

I think things like this are reasons why I don't usually do work on the weekends...

lukeshingles (approved):

PR #315, opened 2026-02-27
Update collisional recombination ratecoefficient to avoid non-finite results

AMG3141:
Small change similar to #303, #307 etc but doesn't exactly fit in with them

### DIFF
--- a/ltepop.h
+++ b/ltepop.h
@@ -32,14 +32,6 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
   return sahafact;
 }
 
-// calculates saha factor in LTE multiplied by exp(-E_threshold / KB / T): Phi_level,ion,element =
-// nn_level,ion,element/(nne*nn_upper,ion+1,element)
-// avoids possible overflows when using the regular sahafact
-[[gnu::const]] [[nodiscard]] constexpr auto calculate_modified_sahafact(const double g_lower, const double g_upper,
-                                                                        const double T) -> double {
-  return SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
-}
-
 [[gnu::pure]] [[nodiscard]] inline DEVICE_FUNC auto get_cellcache_levelpop(const int nonemptymgi,
                                                                            const int uniquelevelindex) -> double {
   assert_testmodeonly(globals::cellcache[cellcacheslotid].nonemptymgi == nonemptymgi);
--- a/ltepop.h
+++ b/ltepop.h
@@ -37,10 +37,7 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
 // avoids possible overflows when using the regular sahafact
 [[gnu::const]] [[nodiscard]] constexpr auto calculate_modified_sahafact(const double g_lower, const double g_upper,
                                                                         const double T) -> double {
-  const double modified_sahafact = SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
-  // TODO: is this assert necessary?
-  assert_testmodeonly(std::isfinite(modified_sahafact));
-  return modified_sahafact;
+  return SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
 }
 
 [[gnu::pure]] [[nodiscard]] inline DEVICE_FUNC auto get_cellcache_levelpop(const int nonemptymgi,
--- a/ltepop.h
+++ b/ltepop.h
@@ -32,6 +32,17 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
   return sahafact;
 }
 
+// calculates saha factor in LTE multiplied by exp(-E_threshold / KB / T): Phi_level,ion,element =
+// nn_level,ion,element/(nne*nn_upper,ion+1,element)
+// avoids possible overflows when using the regular sahafact
+[[gnu::const]] [[nodiscard]] constexpr auto calculate_modified_sahafact(const double g_lower, const double g_upper,
+                                                                        const double T) -> double {
+  const double modified_sahafact = SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
+  // TODO: is this assert necessary?
+  assert_testmodeonly(std::isfinite(modified_sahafact));
+  return modified_sahafact;
+}
+
 [[gnu::pure]] [[nodiscard]] inline DEVICE_FUNC auto get_cellcache_levelpop(const int nonemptymgi,
                                                                            const int uniquelevelindex) -> double {
   assert_testmodeonly(globals::cellcache[cellcacheslotid].nonemptymgi == nonemptymgi);
--- a/macroatom.cc
+++ b/macroatom.cc
@@ -649,7 +649,7 @@ void macroatom_close_file() {
       const double statw_upper = stat_weight(element, upperion, upper);
 
       // Expanded and simplified from
-      // nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * std::exp(-fac1) / fac1
+      // nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * exp(-E/KB/T_e) / (E/KB/T_e)
       const double C =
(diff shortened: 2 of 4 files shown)


lukeshingles:
What kind of testing have you done with this? I guess the fact that none of the checksums changed means that we're completely insensitive to this rate coefficient in the test cases.

ltepop.h
```diff
@@ -32,6 +32,17 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
...
 }
 
+// calculates saha factor in LTE multiplied by exp(-E_threshold / KB / T): Phi_level,ion,element =
+// nn_level,ion,element/(nne*nn_upper,ion+1,element)
+// avoids possible overflows when using the regular sahafact
+[[gnu::const]] [[nodiscard]] constexpr auto calculate_modified_sahafact(const double g_lower, const double g_upper,
+                                                                        const double T) -> double {
+  const double modified_sahafact = SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
+  // TODO: is this assert necessary?
+  assert_testmodeonly(std::isfinite(modified_sahafact));
```
lukeshingles: Not really necessary. If g_upper is zero, this should be caught already when the atomic data is read in.

AMG3141:
> What kind of testing have you done with this? I guess the fact that none of the checksums changed means that we're completely insensitive to this rate coefficient in the test cases.

And of course as soon as you say that one of them fails🙃.

AMG3141:
When you expand out the saha factor its exponential cancels out exactly with the other one in the expression, then I've just inlined the fac1 variable.

lukeshingles:
> > What kind of testing have you done with this? I guess the fact that none of the checksums changed means that we're completely insensitive to this rate coefficient in the test cases.
> 
> And of course as soon as you say that one of them fails🙃.

That's a good thing. If you're rearranging expressions, the tests should fail. Have you compared the new and old numbers side by side? Just the first few is all you need, because if you miss a factor somewhere, all of them will be wrong.

AMG3141:
> > > What kind of testing have you done with this? I guess the fact that none of the checksums changed means that we're completely insensitive to this rate coefficient in the test cases.
> > 
> > 
> > And of course as soon as you say that one of them fails🙃.
> 
> That's a good thing. If you're rearranging expressions, the tests should fail. Have you compared the new and old numbers side by side? Just the first few is all you need, because if you miss a factor somewhere, all of them will be wrong.

I'll go do that now

ltepop.h
```diff
@@ -32,6 +32,14 @@ void set_groundlevelpops(int nonemptymgi, int element, float nne, bool force_sah
   return sahafact;
 }
 
+// calculates saha factor in LTE multiplied by exp(-E_threshold / KB / T): Phi_level,ion,element =
+// nn_level,ion,element/(nne*nn_upper,ion+1,element)
+// avoids possible overflows when using the regular sahafact
+[[gnu::const]] [[nodiscard]] constexpr auto calculate_modified_sahafact(const double g_lower, const double g_upper,
+                                                                        const double T) -> double {
+  return SAHACONST * g_lower / g_upper * std::pow(T, -1.5);
+}
```
lukeshingles: I don't think it's worth defining a separate function for such a simple expression. Just write this into the definition of modified_sahafact in col_recombination_ratecoeff()

lukeshingles (changes requested):

macroatom.cc
```diff
@@ -643,17 +643,16 @@ void macroatom_close_file() {
...
 
       const double sigma_bf = (get_phixs_table(lowerionlower_uniquelevelindex)[0] *
                                get_phixsprobability(lowerionlower_uniquelevelindex, phixstargetindex));
       const double statw_upper = stat_weight(element, upperion, upper);
-      const double sf = calculate_sahafact(statw_lower, statw_upper, T_e, epsilon_trans);
+      const double modified_sahafact = calculate_modified_sahafact(statw_lower, statw_upper, T_e);
 
-      const double C = nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * std::exp(-fac1) / fac1;
+      const double C =
+          nne * nne * modified_sahafact * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * KB * T_e / epsilon_trans;
```
lukeshingles: You can combine the std:pow(T_e, -0.5) and T_e factors into std::sqrt(T_e).

lukeshingles: Actually, it will cancel further with the T_e the modified Saha factor. I think it ends up as just 1 / T_e, but please check this.

globals.h
AMG3141, line 1: Changes from develop

input.cc
AMG3141, line 1: Changes from develop

lukeshingles: These changes from recent develop commits shouldn't be showing up in your PR. What command are you using to rebase? If my branch has 5 of my commits on it, I would normally do this:
```sh
git rebase HEAD~5 --onto develop
git push -f
```

macroatom.cc
```diff
@@ -641,17 +643,17 @@ void macroatom_close_file() {
     if (get_phixsupperlevel(lowerionlower_uniquelevelindex, phixstargetindex) == upper) {
       // Seaton approximation: Mihalas (1978), eq.5-79, p.134
 
-      const double fac1 = epsilon_trans / KB / T_e;
-
```
AMG3141, line 645: Changes from this branch

macroatom.cc
```diff
@@ -641,17 +643,17 @@ void macroatom_close_file() {
...
       const double sigma_bf = (get_phixs_table(lowerionlower_uniquelevelindex)[0] *
                                get_phixsprobability(lowerionlower_uniquelevelindex, phixstargetindex));
       const double statw_upper = stat_weight(element, upperion, upper);
-      const double sf = calculate_sahafact(statw_lower, statw_upper, T_e, epsilon_trans);
 
-      const double C = nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * std::exp(-fac1) / fac1;
+      // Expanded and simplified from
+      // nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * std::exp(-fac1) / fac1
+      const double C =
+          nne * nne * SAHACONST * statw_lower / statw_upper * 1.55e13 * g * sigma_bf * KB / T_e / epsilon_trans;
```
AMG3141, line 654: Changes from this branch

update_packets.cc
AMG3141, line 1: Changes from develop

input.cc
AMG3141: I normally use `git rebase develop` then force push, it's not something I've had to do much in the past so not super familiar with the details of it all.

lukeshingles: Somehow when you rebase, the recent develop commits are being modified (adding yourself as a coauthor). Hold up while I rebase your branch. The checksums will need another update too since these were changed by #320

lukeshingles: Ok, rebased. I did git rebase `HEAD~7 --onto develop` and then `git rebase --skip` twice to skip the checksum updates (which need to be redone anyway). Assuming you've checked that col_recombination_ratecoeff() still returns the same numbers as before, then we just need one more checksum update and we can merge this.

macroatom.cc
```diff
@@ -641,17 +641,17 @@ void macroatom_close_file() {
...
       const double g = gaunt_factor(get_ionstage(element, upperion - 1));
 
       const double sigma_bf = (get_phixs_table(lowerionlower_uniquelevelindex)[0] *
                                get_phixsprobability(lowerionlower_uniquelevelindex, phixstargetindex));
       const double statw_upper = stat_weight(element, upperion, upper);
-      const double sf = calculate_sahafact(statw_lower, statw_upper, T_e, epsilon_trans);
 
-      const double C = nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * std::exp(-fac1) / fac1;
+      // Expanded and simplified from
+      // nne * nne * sf * 1.55e13 * std::pow(T_e, -0.5) * g * sigma_bf * std::exp(-fac1) / fac1
```
lukeshingles: fac1 is no longer defined anywhere. Just substitute in the definition exp(-E/[kT]) / (E/[kT])

lukeshingles (approved):

PR #317, opened 2026-02-27
Don't read or write ratecoeff.dat and calculate integrals at startup

lukeshingles:
With the new Boost integrator, the time to read the file is comparable to the time to calculate the integrals each time. This removes a lot of complexity and a ~100 MB ratecoeff.dat file.


lukeshingles:
@copilot update the summary

PR #375, opened 2026-05-14
Update NLTE pop matrix normalisation to use Iterative Osborne Balancing

AMG3141:


nltepop.cc
```diff
@@ -708,37 +708,57 @@ void nltepop_matrix_add_autoionisation(const int nonemptymgi, const int element,
   }
 }
 
-void nltepop_matrix_normalise(const int nonemptymgi, const int element, std::span<double> rate_matrix,
-                              std::span<double> pop_normfactors, const int first_ion_used, const int nions_used) {
+// Calculate the l2 norm of row or column `i` of `n*n` matrix
+auto norm(std::span<double> matrix, const int i, const bool row, const size_t n) -> double {
```
lukeshingles: Can make matrix a `std::span<const double>` type.
Rename to something a bit more descriptive like 'matrix_row_or_col_norm'

nltepop.cc
```diff
@@ -708,37 +708,57 @@ void nltepop_matrix_add_autoionisation(const int nonemptymgi, const int element,
...
+  for (size_t j = 0; j < n; j++) {
+    const double val = row ? matrix[(i * n) + j] : matrix[(j * n) + i];
+    result += val * val;
+  }
+
+  return std::sqrt(result);
+}
+
+// Multiply row or column `i` by `val`
+void mult(std::span<double> matrix, const int i, const bool row, const size_t n, const double val) {
```
lukeshingles: 'mult' is not very self-explanatory. I suggest something like 'matrix_scale_row_or_col'

nltepop.cc
```diff
@@ -708,37 +708,57 @@ void nltepop_matrix_add_autoionisation(const int nonemptymgi, const int element,
...
+      }
 
-    if (level_isinsuperlevel(element, ion, level)) {
-      // levels in the superlevel get combined together
-      for (int dummylevel = level + 1; dummylevel < get_nlevels(element, ion); dummylevel++) {
-        if (level_isinsuperlevel(element, ion, dummylevel)) {
-          pop_normfactors[column] += calculate_levelpop_boltzmann(nonemptymgi, element, ion, dummylevel);
-        }
+      const double f = std::sqrt(col_norm / row_norm);
+      if (std::abs(f - 1.0F) > 1e-3) {
```
lukeshingles: Why is a 1.0 an f32 here?

lukeshingles (changes requested):
I have a couple of minor suggestions but this looks good overall. Download and extract checksums.zip from the latest CI run and update.

I'll wait to merge this until I've checked a full-scale W7 run.

AMG3141:
Old vs new nlte_0001.out and nlte_0002.out files from local runs of nltephotospheric_dynamic_ion_range_1d_1dgrid.

nlte_0001.out
image

nlte_0002.out
image

These are the main differences between the two, where in the first case the new version sets the entire population of Iron III to 0, and in the second case the old version sets the entire population of Iron I to 0.

nltepop.cc
AMG3141: Mistake, I'll change it to a double

lukeshingles (approved):

lukeshingles:
Queue time on Virgo is long for the W7 results. Let's merge this after the checks pass, and if we find a problem later on, we can follow up or revert.

PR #403, opened 2026-06-08
Improve boundary accuracy

lukeshingles:


grid.cc
```diff
@@ -2332,24 +2261,61 @@ auto get_totmassnuclide_tmodel(const int z, const int a) -> double { return totm
...
+      boundaryposition = std::nextafter(boundaryposition, moved_up ? std::numeric_limits<double>::infinity()
+                                                                   : -std::numeric_limits<double>::infinity());
+      if (prop_gridtype == GridType::CYLINDRICAL2D) {
+        if (d == 1) {
+          // z coordinate is a Cartesian position component
+          pkt.pos[2] = boundaryposition;
+        }
+        // the cylindrical-radius boundary (d == 0) is solved robustly via the quadratic intersection and is
+        // not subject to the Cartesian-plane fragility, so it is left unsnapped
+      } else if (prop_gridtype == GridType::CARTESIAN3D) {
```
lukeshingles, line 2310: The snap (and its `nextafter` nudge) is only applied to Cartesian axes and the cylindrical-z component; the cylindrical-radius (`d == 0`) and all SPHERICAL1D crossings are left unsnapped, with `pkt.cellindex = next_cellindex` but `pkt.pos` kept at the raw `move_pkt_withtime` value.

That means the "membership holds by construction" guarantee in the `change_cell` comment is only enforced for those Cartesian-plane faces. For radial crossings the code still relies on the quadratic-intersection distance landing the position on the correct side of the boundary — i.e. exactly the FP-cancellation fragility this PR is removing elsewhere. The testmode invariant `get_cellindex_from_pos(pkt.pos, pkt.prop_time) == pkt.cellindex` (update_packets.cc:126/173/182, packet.cc:72) therefore still depends on that cancellation being exact for spherical/cyl-radius grids.

Is the quadratic solution guaranteed to land inside the destination cell under `-ffast-math`, or would snapping the radial coordinate too (nudged like the Cartesian case) make the invariant robust for all grid types? Worth a sentence in the comment either way so a future reader knows it's a deliberate choice rather than an omission.

---

grid.cc
```diff
@@ -1227,44 +1185,13 @@ constexpr auto get_grid_type_name(const GridType gridtype) -> std::string {
...
-      }
-      break;
-    }
-  }
-
-  assert_always(false);
-  return -1;
+  // coordinate bins are closed-lower, so we want the highest index such that coord_bin_lower[axis][index] <= pos
+  const auto binposmin = coord_pos_min_tmin[axis] |
+                         std::views::transform([time](double binlower) { return binlower / globals::tmin * time; });
```
lukeshingles, line 1191: Minor (hot path): the `transform` divides every probed bin by `globals::tmin` and multiplies by `time` inside the `upper_bound` binary search, so the loop-invariant `time / globals::tmin` factor is recomputed at each of the ~log2(ncoordgrid) probes. You could compare the raw tmin-frame bins against a single pre-scaled `pos`:

```cpp
const double pos_tmin = pos * globals::tmin / time;
const auto idx = static_cast<int>(
    std::ranges::upper_bound(coord_pos_min_tmin[axis], pos_tmin) - coord_pos_min_tmin[axis].begin()) - 1;
```

This drops the per-probe division/multiply and the `views::transform` wrapper, while giving the same closed-lower result.

---

lukeshingles (commented):
Automated review of the grid coordinate refactor + boundary snap. Overall this looks solid — the per-axis `coord_pos_min_tmin` representation is a clean simplification over the old per-cell `propcell_pos_min`, and the `std::nextafter` nudge in `snap_pkt_to_crossed_boundary` correctly handles the closed-lower convention so that both upward and downward Cartesian crossings land inside the destination cell (an earlier version that snapped exactly onto the face would have put downward crossings on the excluded upper edge, i.e. back in the cell being exited — that's now fixed).

Two small notes left inline:

1. **Radial crossings are left unsnapped** (grid.cc:2309) — the snap only covers Cartesian planes and cylindrical-z, so the "membership by construction" guarantee doesn't extend to spherical / cylindrical-radius crossings, which still rely on exact FP cancellation for the testmode invariant. Likely intentional, but worth documenting (or snapping the radial coord too).
2. **Hot-path micro-opt** (grid.cc:1190) — `get_poscoordpointnum` recomputes the loop-invariant `time/tmin` scale inside each binary-search probe; pre-scaling `pos` once avoids it.

Neither is blocking. Also confirmed the shared-face exactness is preserved despite dropping the old "read the neighbour's min" logic in `boundary_distance`, since both sides now read the same `coord_pos_min_tmin[axis][idx+1]` element; and the removal of the `-99` return from `get_cellindex_from_pos` is benign (its one consumer in the `isoutside_error` block is gated by an unconditional `assert_always(!isoutside_error)`).

---

lukeshingles:
Closed in favour of #404

PR #474, opened 2026-07-28
Fix defects in log messages

lukeshingles:
- nltepop.cc: swapped format arguments printed the ion name in the level
  slot and the level number in the ion-name slot; also remove an
  unbalanced closing parenthesis
- nonthermal.cc: wrong units on the non-thermal deposition rate
  (eV/cm/s/cm^3 -> eV/s/cm^3) and a doubled space; drop an iteration
  count that always printed 10 (the refinement loop has no early exit)
- radfield.cc: the full-spectrum fit line labelled a wavelength as
  "nubar ... Angstrom"; rename to lambda_bar and add kelvin units
- ratecoeff.cc: printlog inside an OpenMP parallel for wrote to the
  threadprivate (unopened) output stream of worker threads, dropping and
  garbling the ion stage list; build the list first and print one line
- vpkt.cc: format spec {:1f} is width 1 (prints 6 decimals), not
  precision 1; use {:.1f}
- input.cc: the photoionisation probability-sum error now ends its line
  and identifies the level; failed input.txt renames are now detected
  and reported instead of silently ignored
- gammapkt.cc: report failures of the legacy gamma line list renames
  instead of logging success before an unchecked std::rename
- grid.cc: fix "READIN GRID SNAPSHOT" wording; sn3d.cc: fix "mean mean"
  typo

No changes to simulation results.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01EijzCW3VtghGPtvT3MQjrk

### DIFF
--- a/decay.cc
+++ b/decay.cc
@@ -103,8 +103,8 @@ std::vector<Nuclide> nuclides;
 std::vector<DecayPath> decaypaths;
 std::vector<bool> alldecaytypes_is_used;
 
-[[nodiscard]] constexpr auto decay_daughters_z_a_prob(const int z_parent, const int a_parent, const DecayType decaytype)
-    -> std::vector<DecayDaughter> {
+[[nodiscard]] constexpr auto decay_daughters_z_a_prob(const int z_parent, const int a_parent,
+                                                      const DecayType decaytype) -> std::vector<DecayDaughter> {
   assert_always(decaytype >= 0);
   assert_always(decaytype < DecayType::DECAYTYPE_COUNT);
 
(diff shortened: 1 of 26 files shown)


lukeshingles:
CI status: the only red check is **NVIDIA NVHPC nvc++**, and it is not caused by this branch.

The job fails during setup, before it compiles anything:

```
Err:1 https://developer.download.nvidia.com/hpc-sdk/ubuntu/amd64  nvhpc-26-1 26.1-0
  404  Not Found [IP: 23.32.152.140 443]
E: Failed to fetch https://developer.download.nvidia.com/hpc-sdk/ubuntu/amd64/./nvhpc-26-1_26.1-0_amd64.deb  404  Not Found
```

The same job is the only failure on `develop` too, including this PR's base commit `8c29ff1`. It first went red on `336a66c` (27 Jul), before this branch existed; the last green `develop` run was `2e01666`.

Everything else is green on the current head `a8e7098`: Ubuntu gcc 15, gcc 16, clang 22, Apple Clang and LLVM Clang, cppcheck, and all regression tests in both the normal and `TESTMODE=ON` workflows, with `combine_checksums` passing. The stored md5 checksums are therefore unchanged, which is the expected result for these log-output-only changes.

No action needed on the branch for this. I'll re-check and re-run the job once NVIDIA publishes the `nvhpc-26-1` package again.

---

PR #488, opened 2026-07-31
Make the NLTE solver reject non-finite solutions in both linear algebra backends and judge convergence by backward error

lukeshingles:
## Context

`nltepop_matrix_solve` carries an `#ifdef EIGEN_OFF` split between GSL (`make EIGEN=OFF`) and Eigen (the default). #486 fixed three divergences where the two builds *crashed* differently on the same input, and listed the remaining known ones. CI compiles `EIGEN=OFF` but never runs it, so the GSL path has no regression coverage.

Re-auditing the two branches turned up a wrong-answer divergence, not just a diagnostic one, plus two backend-independent numerical defects downstream of it. This PR fixes all three, makes the two refinement branches read line-for-line the same, and replaces a meaningless convergence diagnostic with a well-defined one.

## Defect 1: the error metric was a different function in each branch, and neither was NaN-safe

The refinement loop decides whether a solve succeeded from `error = max|residual|`. Measured directly against gslcblas 2.7.1 and the vendored Eigen:

| residual vector | GSL `idamax` | Eigen `maxCoeff` |
|---|---|---|
| NaN at index 0, finite elsewhere | `3e-50` | `nan` |
| NaN in the middle | `3e-50` | `3e-50` |
| NaN at the end | `3e-50` | `3e-50` |
| all NaN | `nan` | `nan` |
| `+inf` in the middle | `inf` | `inf` |
| all finite (control) | `5.5` | `5.5` |

gslcblas's `idamax` selects with a bare `>` comparison, which a NaN never satisfies. Eigen's `maxCoeff` defaults to `PropagateFast`, documented as undefined for NaN (its scalar and SIMD paths genuinely differ; the scalar path propagates only a NaN in the first element).

Index 0 is exactly where this solver puts it: row 0 is the dense population normalisation constraint, so a NaN anywhere in the solution vector poisons `residual[0]`. On that input the Eigen build reported a non-finite `error`, never latched the garbage, and returned false — letting the caller drop an ion stage and retry — while the GSL build got a small finite `error` from one of the many rows the sparse matrix leaves finite, latched the garbage solution, satisfied the `error < 1e-40` early exit and **accepted it**. The Eigen build's rejection was incidental, not guaranteed: it depended on the NaN's position and on a reduction order Eigen documents as undefined. Neither build was safe.

**Fix:** `error` is computed once, outside the backend switch, by a shared scalar helper that propagates a non-finite element from any position (Eigen provides `maxCoeff<PropagateNaN>`, but GSL has no counterpart, and one shared function guarantees the builds agree). For an all-finite residual it returns exactly what each backend returned before. A `#error` guard rejects `-ffinite-math-only` builds, which would silently disable the entire non-finite detection scheme.

## Defect 2: a non-finite population was laundered into MINPOP and reported as success

Under `STRICT_POPULATION_CHECKING`, every comparison in `solution_pops_are_valid`'s triage is false for a NaN: the `MINPOP` ground check, the `-MINPOP` negative check, and both inversion factors. NaN and `+inf` fell through to the "likely a rounding error to zero" branch, were overwritten with `MINPOP`, and the solve was reported successful — which then satisfies the `assert_always(std::isfinite(pop))` in `nltepop_apply_solution` rather than tripping it. Only `-inf` was handled correctly.

**Fix:** a non-finite population — reachable via the denormalisation product `vec_x[i] * pop_normfactors[i]` overflowing, since the solver itself now rejects non-finite solution vectors through the residual score — fails the solve up front, and `popvec` is zero-filled on that failure so `can_remove_ion`'s retry triage reads zeros (as after a singular-matrix failure) rather than non-finite values whose comparisons are meaningless. The negative branch keeps its rounding-error logic for the finite values it was written for.

## Defect 3: the convergence warning compared a dimensional residual against a fixed constant

`nltepop_matrix_normalise` rescales each row by an arbitrary accumulated balancing factor, so the residual elements carry no fixed physical magnitude, and the old absolute `error_best > 1e-8` test compared that against a pure number. Measured on `nltephotospheric_dynamic_ion_range_1d_1dgrid` (n_Fe ~ 1e11 cm^-3), rank 0 alone emitted **45** such warnings under Eigen and **46** under GSL, all on healthy solves — and every log line is flushed. On `nebular_1d_3dgrid` it never fired: the absolute threshold either spams or stays silent depending only on density. Its printed iteration count was also off by one on early exit and never identified which pass produced the accepted solution.

**Fix:** convergence is judged by the **componentwise relative backward error**, `max_i |b_i − (A·x)_i| / (|b_i| + Σ_j|A_ij·x_j|)`, in one O(n²) pass over the equilibrated system. Per-row rather than normwise is load-bearing: the equilibration balances each row's norm against its column's, never rows against each other, and with `FORCE_SAHA_ION_BALANCE` the constraint rows carry right-hand sides differing by many orders of magnitude, so a global ratio would let a badly violated small-scale row hide behind the largest row; the arbitrary per-row factors cancel in the ratio. The warning names the refinement pass that produced the best solution (pass 1 being the unrefined LU solution).

## The refinement loop now reads the same in both branches

The GSL branch called `gsl_linalg_LU_refine` (which forms the residual internally) and then formed `A·x − b` *again* with its own GEMV, while the Eigen branch used `b − A·x` under the same variable name. Both branches now read `correction = A⁻¹·residual; x += correction; residual = b − A·x`, with the `#ifdef` covering only the backend calls (`LU_svx` + `daxpy` on the GSL side) and the error scored below it. This drops one of the two O(n²) passes the GSL build did per refinement iteration, and the shared named correction buffer avoids the evaluator temporary Eigen heap-allocated for `x += lu.solve(r)` on every iteration (the buffer is a symmetry choice — Eigen handles an aliased solve destination correctly). Also hoisted `gsl_linalg_LU_decomp` out of the `assert_always` that was performing the decomposition as a side effect.

## Results are unchanged

Baseline-vs-changed regression runs on one machine, `REPRODUCIBLE=ON FASTMATH=OFF MAX_NODE_SIZE=2`, against the same `origin/develop` baseline, at the head commit:

| Test | Eigen | `EIGEN=OFF` |
|---|---|---|
| `nebular_1d_3dgrid` | 26/26 identical | 26/26 identical |
| `nltephotospheric_dynamic_ion_range_1d_1dgrid` | 27/27 identical | 27/27 identical |

The baseline Eigen run of `nebular_1d_3dgrid` reproduces the stored CI checksums exactly, so the comparison is against the same numbers CI checks. (In the photospheric test three files — one `nlte_*.out`, two `radfield_*.out` — differ from the stored checksums both before and after the change: a pre-existing g++-14 vs g++-15 difference, not one introduced here.)

The `EIGEN=OFF` results being bit-identical is the empirical backing for the claim that replacing `gsl_linalg_LU_refine` with `LU_svx` + `daxpy`, and flipping the GEMV's alpha/beta, is an exact negation at every step. The componentwise measure — strictly more sensitive than a normwise one — fires zero warnings on healthy solves of both models under both backends.

Also verified:
- All four of {Eigen, `EIGEN=OFF`} × {default, `TESTMODE=ON`} compile clean under `-Werror` (the `TESTMODE=ON` builds matter, since `assert_testmodeonly` bodies are not compiled otherwise).
- A `TESTMODE=ON EIGEN=OFF` run completes with no assertion or sanitizer failure.
- Unit tests pass for the classic (53/53) and nebular (52/52) presets.
- clang-format is clean and clang-tidy reports no diagnostics beyond the three pre-existing ones.

## Deliberately out of scope

`sfmatrix_solve` in `nonthermal.cc` has the identical NaN-blind error metric, where the consequence is a failed `assert_always(error_best >= 0.)` rather than a corrupted population — left for a separate change. Also untouched: the `error < 1e-40` early exit, which is effectively dead but cannot become a usable relative tolerance without truncating the best-of-ten selection and changing the solution; the fixed count of ten refinement iterations; in-place GSL solving (would reintroduce the branch asymmetry this PR removes); and `gsl_set_error_handler_off()`, which this solver does not need because `lumatrix_is_singular` already rejects the only input that would make GSL abort.

https://claude.ai/code/session_01MsXwyCR7wwnCgnxvCqzq4C

### DIFF
--- a/nltepop.cc
+++ b/nltepop.cc
@@ -839,16 +839,19 @@ void set_element_pops_lte(const int nonemptymgi, const int element) {
         popvec[index] = ltepop;
       }
     } else {
-      // A non-finite population is unrecoverable, and has to be rejected before the triage below, because every
-      // comparison there is false for a NaN: the MINPOP ground check, the -MINPOP negative check, and the inversion
-      // factors. Without this, NaN and +inf reach the "rounding error to zero" branch, are replaced by MINPOP, and
-      // the solve is reported as successful - which then satisfies the assert_always(std::isfinite(pop)) in
-      // nltepop_apply_solution instead of tripping it.
+      // A non-finite population (reachable via the popvec[i] = vec_x[i] * pop_normfactors[i] denormalisation
+      // product overflowing - the solver itself rejects a non-finite solution vector through the residual score)
+      // fails every comparison in the triage below, so without this rejection it would pass through unhandled and
+      // trip the assert_always(std::isfinite(pop)) in nltepop_apply_solution, aborting the run instead of letting
+      // the caller drop an ion stage and retry.
       if (!std::isfinite(population)) {
         printlnlog(
             "  [warning] cell {} ts {}: NLTE solver gave non-finite population for index {} (Z={} ionstage {} level "
             "{}), pop = {:g}. Returning nltepop_matrix_solve failure",
             nltelog.modelgridindex, nltelog.timestep, index, get_atomicnumber(element), ionstage, level, population);
+        // clear the failed attempt so that can_remove_ion's triage reads zeros, as it does after a singular-matrix
+        // failure, rather than non-finite values whose comparisons would all be meaningless
+        std::ranges::fill(popvec, 0.);
         return false;
       }
 
@@ -964,15 +967,21 @@ template <typename GetDiagElement>
   return is_singular;
 }
 
-// Largest absolute element of a vector, propagating non-finite elements: as soon as one is found, its magnitude
-// (inf or nan) is returned instead of the largest finite magnitude.
-// Neither backend's own reduction can be used for this, because neither propagates a NaN reliably and the two drop
-// them in different places, so the same residual can be scored differently by the two builds. gslcblas's idamax
-// seeds its running maximum at zero and selects with a bare `>`, so it skips a NaN in any position and only returns
-// one when every element is NaN. Eigen's maxCoeff defaults to PropagateFast, which Eigen documents as giving an
-// undefined result for NaN; in practice it keeps the running maximum on the left of the comparison, so it returns a
-// NaN only when the first element is one. Either way a solution vector that is entirely garbage can be scored with a
-// small finite residual and accepted. (An infinity does propagate through both.)
+// The non-finite detection in this solver relies on std::isfinite and NaN comparisons keeping their IEEE semantics,
+// which the FASTMATH build preserves only because -fno-finite-math-only is appended after -ffast-math (Makefile).
+#if defined(__FINITE_MATH_ONLY__) && __FINITE_MATH_ONLY__
+#error \
+    "the NLTE solver requires std::isfinite to detect NaN/inf: compile without -ffinite-math-only (see FASTMATH in the Makefile)"
+#endif
+
+// Largest absolute element of a vector, except that the magnitude of the first non-finite element (inf or nan) is
+// returned as soon as one is found - so callers must test the result only with std::isfinite, never distinguishing
+// inf from nan. Neither linear algebra backend's own reduction is usable here, because neither propagates a NaN
+// reliably and the two drop them in different positions, so the same residual could be scored differently by the two
+// builds: gslcblas's idamax selects with a bare `>` comparison, which a NaN never satisfies, and Eigen's default
+// maxCoeff uses PropagateFast, which Eigen documents as undefined for NaN (its scalar and SIMD paths genuinely
+// differ). Eigen does provide maxCoeff<PropagateNaN>, but GSL has no counterpart, and scoring through one shared
+// scalar function guarantees the two builds agree.
 [[nodiscard]] auto max_abs_propagate_nonfinite(const std::span<const double> values) -> double {
   double maxabs = 0.;
   for (const double value : values) {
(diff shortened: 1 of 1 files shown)


lukeshingles:
A multi-angle review of this PR's two commits (eleven independent verification passes over the diff, the vendored Eigen, and the installed GSL) produced one substantive correction plus a round of comment fixes, pushed as `315737e`.

**The substantive fix: the relative-residual warning now uses the componentwise backward error.** The normwise measure introduced earlier (`max|r| / max_i(|b_i| + Σ|A_ij·x_j|)`) could hide a badly violated row: the equilibration balances each row's norm against its column's but never rows against each other, and with `FORCE_SAHA_ION_BALANCE` the constraint rows carry right-hand sides differing by many orders of magnitude, so a minor ion's row could be 100% wrong while the global ratio stayed below tolerance. The measure is now `max_i |b_i − (A·x)_i| / (|b_i| + Σ_j|A_ij·x_j|)` in the same single O(n²) pass — the per-row equilibration factors cancel in the ratio, and the previously flagged fallback branch (which printed an absolute value labeled "relative") disappears entirely. A proposed O(n) guard to skip the pass was rejected as unsound for the componentwise measure.

Also in this round: `popvec` is zero-filled when a non-finite population fails the solve, so `can_remove_ion`'s retry triage reads zeros (as after a singular-matrix failure) instead of NaN/inf whose comparisons are meaningless; a `#error` guard now rejects `-ffinite-math-only` builds, which would silently disable the whole non-finite detection scheme; and several comments that the review proved wrong against the sources were corrected (the non-finite guard's counterfactual described the pre-change code; Eigen's solve does *not* alias its destination — the correction buffer is a symmetry choice; the `assert_always` "must not" rule was false; the idamax/maxCoeff NaN characterisations were imprecise, and `maxCoeff<PropagateNaN>` exists — the shared helper is used because GSL has no counterpart).

**Results verified unchanged, again on all four combinations** ({Eigen, `EIGEN=OFF`} × {nebular, photospheric}, `REPRODUCIBLE=ON FASTMATH=OFF MAX_NODE_SIZE=2`, against the same develop baseline): 26/26 and 27/27 checksums identical everywhere, and the componentwise measure — strictly more sensitive than the normwise one — still fires zero warnings on healthy solves of both models. All four build configs compile clean under `-Werror`, unit tests pass (53/53 classic, 52/52 nebular), clang-format clean, clang-tidy unchanged.

Reviewed and deliberately not adopted (reasons in the commit message): sharing the error metric with `sfmatrix_solve` (out of scope, as documented), in-place GSL solving (reintroduces branch asymmetry), a per-backend-lambda loop restructure, and hoisting the non-finite check above the strictness split (patch-and-continue is the non-strict mode's contract).

---

PR #512, opened 2026-08-10
Batch NLTE bound-bound rate coefficients and speed up matrix equilibration

lukeshingles:
## What changed

**Single-pass bound-bound fill.** `nltepop_matrix_add_boundbound()` visited every transition twice — once in the upper level's down-transition loop and once in the lower level's up-transition loop — with each visit independently recomputing the `nu_trans` → `B_ul` → `tau_sobolev` → `expm1` chain, and the collisional pair recomputing related exponentials. A new `calculate_boundbound_ratecoeffs()` in `macroatom.cc` computes all four coefficients of a transition together: the Sobolev optical depth and escape probability are shared between the radiative directions, and collisional excitation is derived from de-excitation via detailed balance (`C_lu = C_ul (g_u/g_l) exp(-epsilon/kT_e)`, which every branch pair of the existing `col_*_ratecoeff()` functions satisfies — the collisional physics still lives only in `col_deexcitation_ratecoeff()`). The fill now drives from the up-transition lists because the non-thermal excitation list is keyed on the uptrans `alltransindex`.

**Batched matrix equilibration.** `nltepop_matrix_normalise()` performed per-index row/column norm and scale passes, the column ones striding through the full matrix, inside an up-to-10-sweep loop. Each sweep now computes all row and column norms in one sequential row-major pass and applies all scalings in a second row-major pass, with the factors taken from the same per-sweep norm snapshot. The scalings are exactly undone via `pop_normfactors`, so they affect only the conditioning of the LU solve, not the system being solved.

## Measured performance

`nebular_1d_3dgrid` setup, timesteps 5–6 resumed from an identical snapshot, M4 Pro, 4 MPI ranks (temporary phase timers, not part of this PR).

**Default build configuration** (`make` defaults: fast-math on, reproducible off, Eigen vectorised), mean per element-solve across 63 solves with identical NLTE iteration counts:

| | main | this PR | change |
|---|---|---|---|
| NLTE matrix fill (all elements) | 6.9 ms | 5.8 ms | **−16%** |
| normalise + LU solve | 2.9 ms | 2.3 ms | **−19%** |
| Fe fill (dim 527) | 10.9 ms | 9.2 ms | −16% |
| Fe normalise + solve | 6.1 ms | 5.0 ms | −18% |
| whole NLTE solver per iteration | 29.2 ms | 24.4 ms | **−16%** |

The equilibration rewrite matters proportionally more here than in reproducible builds because the vectorised LU no longer dominates the solve phase. (A fast-math binary cannot resume a snapshot written by a no-fast-math binary — FMA contraction changes the radfield bin edges and trips the restart consistency check — so the default-configuration comparison used its own snapshot.)

**`REPRODUCIBLE=ON FASTMATH=OFF MAX_NODE_SIZE=2`** (the CI test configuration), phase breakdown for Fe (dimension 527), mean per element-solve across 21 solves:

| phase | before | after |
|---|---|---|
| bound-bound fill | 6.0 ms | 4.5 ms (**−25%**) |
| sum + normalise | 3.7 ms | 2.0 ms (**−46%**) |
| ionisation fill / LU solve | unchanged | unchanged |
| whole NLTE solver per iteration (all elements) | 33.6 ms | 29.2 ms (**−13%**) |

## Results change at roundoff level — checksums need regeneration

The shared-subexpression evaluation and the batched (same-snapshot) equilibration factors alter floating-point rounding paths, not the system being solved. In the benchmark: identical NLTE iteration counts and solver warning counts; cells 0 and 1 byte-identical; the only differences were in the cell with the known near-singular NLTE matrix, at most 2.5e-4 relative and only in a population holding ~1e-9 of its element (noise-dominated in both runs). The nebular test checksums will need regenerating via the "Update checksums" workflow after CI completes.

## Reviewer notes

- The remaining fill time is the adaptive photoionisation rate integrals. A measured experiment with caching the radiation-field-only part per timestep and integrating the stimulated correction over a narrow range gave no net speedup (46% of calls have departure ratio > 1 and must take the clamped path, and the adaptive quadrature cost concentrates near threshold), so they are deliberately untouched.
- 86/86 unit tests pass; clang-format, `OPTIMIZE=OFF`, and full `TESTMODE=ON` builds pass.


lukeshingles:
16% performance gain is not worth the added complexity.

PR #514, opened 2026-08-10
Fill the cumulative bf opacity list only when a continuum event needs it

lukeshingles:
Skip filling the cumulative bound-free opacity list in the propagation hot path, refilling it on demand only when a bound-free absorption event actually needs it. **~10% faster update_packets** at nebular NLTE epochs with **bit-identical outputs** (verified down to the full binary packet state) and **unchanged CI checksums**.

(The companion change that also removes the per-continuum division — another ~12% — is split into a follow-up PR because it perturbs float cross-sections in the last bit on some platforms and needs a checksum regeneration; this PR is exactly result-neutral.)

## Motivation (W7 nebular profile at ts14, per rank)

Counter instrumentation showed the shape of the problem:

- 1.28M propagation steps trigger **1.22M full bf opacity evaluations** (the 1e-4 nu-reuse guard almost never hits because per-step redshift exceeds it), each covering an average window of **13,300 continua** → **16.2 billion loop iterations per rank per timestep**, ~70% of update_packets CPU.
- The `chi_bf_sum` cumulative list that every iteration stores is consumed only when a bf absorption is actually sampled: **1,621 times per rank-timestep (0.13% of evaluations)**.

## Change

The hot path computes the scalar `chi_boundfree` and the estimator contributions only (new `FILLCHIBFSUM` template flag); `rpkt_event_continuum()` refills the list on demand at the frequency of the last evaluation, which reproduces the values that `chi_boundfree` was accumulated from. The refill runs with `FILLESTIMCONTRIBS` off so the estimator contributions (`gamma_contr`, `groundcont_gamma_contr`) keep exactly the values the original evaluation stored — rewriting them could change them in the last bit via cellcache entries that the original evaluation itself populated.

## Measurements (M4 Pro, 4 MPI ranks, default fast-math build)

**W7 nebular, nltenebular preset, MPKTS=1e7, resumed ts14–16** (rank-0 update_packets):

| | ts14 | ts15 | ts16 | total |
|---|---|---|---|---|
| main | 73.6 s | 80.9 s | 73.0 s | 227.5 s |
| this PR | 67.1 s | 68.7 s | 69.5 s | **205.3 s (−10%)** |

Light curves, spectra, deposition, and the full `packets_*.tmp` state after three timesteps are **bit-identical to main**.

**sim2010 classic-mode photospheric model, quasi-NLTE ts1–4, MPKTS=1e6, fixed seed:** performance parity (rank-summed update_packets within 0.3% across repeat runs) and bit-identical outputs — expected, since the grey/thick photospheric regime barely exercises the bf window loop.

## Verification

- CI configuration (`REPRODUCIBLE=ON FASTMATH=OFF MAX_NODE_SIZE=2`): the nebular_1d_3dgrid and nltephotospheric_dynamic_ion_range tests (job0 + job1 resume + exspec) run locally for main and this branch — **all output files md5-identical**, and a commit-1-only probe branch passed the full CI matrix with unchanged checksums.
- 86/86 unit tests pass; classic preset compiles.

### DIFF
--- a/rpkt.cc
+++ b/rpkt.cc
@@ -407,12 +407,15 @@ void electron_scatter_rpkt(Packet& pkt) {
   set_pkt_restframe_from_cmf(pkt);
 }
 
+template <bool USECELLHISTANDUPDATEPHIXSLIST, bool FILLCHIBFSUM = false, bool FILLESTIMCONTRIBS = true>
+auto calculate_chi_bf_gammacontr(int nonemptymgi, double nu, Phixslist& phixslist) -> double;
+
 // Handle a continuum interaction of an r-packet by sampling which continuum process occurs, in
 // proportion to its share of the total continuum opacity: electron scattering (coherent in the comoving
 // frame; see electron_scatter_rpkt()), free-free absorption (packet becomes a k-packet), or bound-free absorption
 // (packet activates a macro-atom in the upper ion with probability nu_edge/nu, the ionisation energy fraction, and
 // otherwise becomes a k-packet carrying the freed electron's kinetic energy).
-void rpkt_event_continuum(Packet& pkt, const ContinuumOpacity& chi_rpkt_cont) {
+void rpkt_event_continuum(Packet& pkt, ContinuumOpacity& chi_rpkt_cont) {
   const double nu = pkt.nu_cmf;
 
   const double dopplerfactor = calculate_doppler_nucmf_on_nurf(pkt.pos, pkt.dir, pkt.prop_time);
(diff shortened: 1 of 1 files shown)


lukeshingles:
The nltephotospheric_dynamic_ion_range checksum failure was a real defect in the first commit: the event-time refill also rewrote the estimator contributions (`gamma_contr`/`groundcont_gamma_contr`), and when the original evaluation had itself populated lazy cellcache entries, the refilled values could differ in the last bit and leak into the bf estimators via later reuse of the cached opacity — packet trajectories were unaffected, which is why only estimator/nlte/radfield files diverged.

Fixed by giving the refill a `FILLESTIMCONTRIBS=false` mode so it fills only the cumulative opacity list and leaves the estimator contributions exactly as the original evaluation stored them. Verified locally in the CI configuration by running this exact test (job0 + job1 resume + exspec) for main and the branch: all output files md5-identical.

_🤖 [Claude Code](https://claude.com/claude-code)_

lukeshingles:
Root cause of the repeated nltephotospheric_dynamic_ion_range failure found and resolved by splitting the PR.

Both failing runs produced bit-identical outputs, so the FILLESTIMCONTRIBS fix wasn't the (whole) story — a probe branch with only the fill-on-demand commit passed the full CI matrix, isolating the checksum change to the division→reciprocal commit. Its last-bit table-position differences occasionally flip a float-rounded cross-section; when the affected continuum is a negligible contributor to the summed opacity, trajectories (packets/spectra/light curves) are unchanged but the per-continuum detailed bf estimators differ at print precision — hence only estimators/nlte/radfield files diverging, deterministically on Linux/glibc but not macOS. 

This PR is now the exactly-result-neutral fill-on-demand change only (−10%). The reciprocal (−12% further) moved to #515, which requires the "Update checksums" workflow for the affected test.

_🤖 [Claude Code](https://claude.com/claude-code)_

PR #518, opened 2026-08-12
Pack the bound-free keep flags into a bitmap and walk only set bits

lukeshingles:
Rebased onto main now that #517 is merged.

## What the loop was doing

`calculate_chi_bf_gammacontr` scans `[allcontbegin, allcontend)` on every opacity evaluation, testing each continuum's keep flag. Instrumenting a throwaway build on sim2010 classic showed how little of that scan is useful:

- **11.3%** of continua are kept in a typical cell
- **57.7%** of 64-continuum blocks contain no kept continuum at all
- mean window is **9324** of 21936 continua

So of ~9324 iterations per evaluation, ~8268 existed only to read a flag and move on.

## What this does

Store the flags as one bit per continuum rather than one byte, and iterate words instead of continua. A zero word clears 64 at a time; within a word, `std::countr_zero` steps straight to the next set bit, so unset bits inside a partly-filled word are skipped too. The loop body runs **~1056 times per evaluation instead of ~9324** — an 8.8× cut in trip count.

Set bits are visited in ascending index order, so the summation order is unchanged and results are bit-identical rather than merely close.

The detailed bound-free estimator entries that skipped continua used to zero individually are now cleared with one `fill` over the bfestim window, which covers exactly the continua of `[allcontbegin, allcontend)` that have an estimator index. That is the only behavioural subtlety here, and it is what the `nebular_1d_3dgrid` regression test exercises.

## Measured — and the gain depends strongly on the configuration

**sim2010 1.06 M☉ classic** (`USE_LUT_PHOTOION`, `DETAILED_BF_ESTIMATORS_ON = false`; 21936 continua, 461k lines, 100k packets, 6 timesteps, 4 ranks, `-O3 -ffast-math`, Apple M4 Pro), 3 order-alternated repeats on an idle machine:

| | before | after | |
|---|---|---|---|
| packet propagation | 73.87 s (sd 0.38) | 29.70 s (sd 0.16) | **2.49×** |
| total wall | 79.20 s | 35.01 s | **2.26×** |

Per-pair propagation ratios agree to within 1%: 0.401, 0.404, 0.402.

**W7 nebular** (`artisoptions_nltenebular.h`, `DETAILED_BF_ESTIMATORS_ON = true`; 47649 continua, 2.84M lines, 1M packets, timesteps 0–7, 4 ranks), 4 order-alternated pairs: summed propagation 14.7 s → 13.8 s. That is within the noise of a ~14 s signal at 0.1 s log resolution, so the honest reading is **no measurable change** — not a regression, but no gain either.

The reason is visible in the same instrumentation: on W7, **100.00% of continua are kept and 0.00% of words are all-zero** (measured over 7980 cell populations). Under `DETAILED_BF_ESTIMATORS_ON` the keep test in `keep_this_cont()` degenerates to `grid::get_elem_massfrac(...) > 0`, i.e. "is this element present in the cell at all", which in W7 is true for every continuum. There is nothing to skip, so the bit-walk reduces to the byte scan plus a little bit-extraction.

**So: a large win for the LUT-photoion configurations (classic, kilonova), neutral for the nebular ones.** If the nebular presets ever want this, the lever is a more selective `keep_this_cont()` under `DETAILED_BF_ESTIMATORS_ON` — but that changes which continua contribute to the opacity, so it is a physics decision rather than a refactor.

Caveats on the W7 numbers: that model directory has no saved state to resume from, so the run covers timesteps 0–7 at 150–160 days rather than the 410-day regime the model is built for, and `num_grey_timesteps = 4` means only four of those timesteps do opacity work at all. `DETAILED_BF_ESTIMATORS_USEFROMTIMESTEP = 13`, so the detailed estimators are accumulated (that is not timestep-gated) but not yet consumed by `get_corrphotoioncoeff()` in the measured range.

Secondary benefit: in multi-slot (GPU) mode the flags are per cell rather than per rank, so this shrinks that allocation 8× — roughly 220 MB to 27 MB for 21936 continua across 10⁴ cells.

## Verification

- Bit-identical to `main` on `kilonova_1d`, `nebular_1d_3dgrid` and `classicmode_1d_3dgrid`, over the full job0 → resume → exspec sequence, plus a `TESTMODE=ON` (assertions + ASan/UBSan) run of the nebular case — the config that exercises the estimator zeroing this restructures. Re-run after the rebase.
- Byte-identical physics output and identical event counters against the pre-change build on both the sim2010 and W7 production runs above.
- All six `artisoptions_*.h` presets compile under clang and gcc-16; `unittests` pass for the classic and nebular presets; clang-format clean.

Results are unchanged, so the stored checksums should not need regenerating.

### DIFF
--- a/constants.h
+++ b/constants.h
@@ -5,7 +5,9 @@
 #define CONSTANTS_H
 
 #include <array>
+#include <bit>
 #include <cstddef>
+#include <cstdint>
 #include <new>
 #include <numbers>
 #include <string_view>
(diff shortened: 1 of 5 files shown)


lukeshingles:
@codex review

PR #540, opened 2026-08-23
Clear the NLTE level populations before the Saha populations of a grey cell

lukeshingles:
## Summary

A cell that had an NLTE solution can go to the grey/LTE path in a later timestep, e.g. when no packet entered it (`W == 1`). That path computed the partition functions before it cleared the NLTE solution, and `nltepop_reset_solution_ranges()` cleared only the ion ranges. The stale excited populations stayed in the cell. The next grey timestep divided them by a Saha ground population at the `MINPOP` floor, and the partition function overflowed a `float`. The assertion in `calculate_partfunct()` then stopped the run.

Found with a new 2D kilonova test with time-dependent NLTE (follow-up PR), where low density outer cells get no packets in some timesteps.

## Change

- `nltepop_reset_solution_ranges()` becomes `nltepop_reset_cell()`. It writes the `-1` marker into the NLTE level populations of every NLTE element (with `nltepop_reset_element()`), which also clears the ion ranges and the time-dependent solution time.
- The grey path in `update_grid_cell()` calls it before the partition functions and the Saha balance.

## Results

The results change only for a cell that goes from an NLTE solution to the grey path. The stored checksums may change if a test has such a cell. Local check of `nebular_1d_3dgrid` on macOS: see the comment below.

### DIFF
--- a/nltepop.cc
+++ b/nltepop.cc
@@ -1632,11 +1632,18 @@ void nltepop_apply_solution(const int element, const int nonemptymgi, const int 
 
 }  // anonymous namespace
 
-// Clear the stored NLTE solution ranges of every element in the cell, for the paths that replace
-// the NLTE solution with Saha populations without a solve.
-void nltepop_reset_solution_ranges(const int nonemptymgi) {
+// Clear the NLTE solution of every element in the cell, for the paths that replace the NLTE
+// solution with Saha populations without a solve. The level populations get the -1 marker, so
+// the partition functions and the Saha balance then use the Boltzmann populations. A stale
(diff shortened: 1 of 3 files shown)


lukeshingles:
Local check on macOS (clang, REPRODUCIBLE=ON MAX_NODE_SIZE=2 FASTMATH=OFF): the checksums of nebular_1d_3dgrid are identical to main for both sets (26 files each). No cell of that test goes from an NLTE solution to the grey path.

lukeshingles:
Folded into #539 at the maintainer's request.

PR #541, opened 2026-08-23
Add the CI test kilonova_2d_timedepnlte

lukeshingles:
## Summary

A new CI test `kilonova_2d_timedepnlte`. It starts from the model of `kilonova_2d` and runs from 7 to 10 days with:

- the nebular NLTE preset (`artisoptions_nltenebular.h`) with 30 excited NLTE levels per ion, `MINTEMP` 1000 K, `MAXTEMP` 20000 K, and the kilonova settings `USE_CALCULATED_MEANATOMICWEIGHT` and the time-dependent particle thermalisation;
- `NLTE_TIME_DEPENDENT_FIRST_TIMESTEP = 3` (#536);
- `NLTE_OUTER_ANDERSON_DEPTH = 2` and `NLTE_OUTER_RELTOL = 0.01` (#539);
- the atomic data set `atomicdata_sryzrlace.tar.zst` (Sr, Y, Zr, La, Ce) from the v2026.5.15 release. `ci.yml` caches it like the other data sets, and the setup script takes `compositiondata.txt` from the archive.

The run has 8 timesteps, 200000 packets, and 116 non-empty cells. Timesteps 0 and 1 are LTE, timestep 2 is a steady-state NLTE solve, and timesteps 3 to 7 are time-dependent. The first job runs timesteps 0 to 3 and the resume starts at timestep 3. The kilonova spectrum plot of this test uses 8.5 days.

This PR stacks on #539 and #540. It includes their commits until they merge. The test found the bug that #540 corrects: low density outer cells get no packets in some timesteps and go to the grey path.

## Local run (macOS, clang, 4 ranks)

- job0 55 s, job1 2 min 20 s, exspec 7 s. The checksum sets have 529 and 531 files.
- Only the five elements with atomic data can cool, but the non-thermal deposition of the whole mass heats them. The temperatures are therefore higher than in a real kilonova, and 44 of 116 cells reach `MAXTEMP` by timestep 7. The test is a regression test of the code paths, not a physics validation.
- One near-vacuum cell (`nne` of 22 cm^-3) has a 2-cycle between a Saha state at `MINTEMP` and an NLTE state at 2000 K, and reaches `NLTEITER` in every time-dependent timestep. This costs about 12 s per timestep.
- No stored checksums yet. The local checksums are from macOS and do not match Linux. A maintainer can make them with the "Update checksums" workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
                         kilonova_2d,
                         kilonova_2d_barnesthermalisation,
                         kilonova_2d_expansionopac,
+                        kilonova_2d_timedepnlte,
                         kilonova_2d_xcomgammaphotoion,
                         nebular_1d_3dgrid,
                         nebular_1d_3dgrid_limitbfest,
@@ -69,7 +70,7 @@ jobs:
                   echo "count=$(python3 -c 'import multiprocessing; print(multiprocessing.cpu_count())')" >> $GITHUB_OUTPUT
 
             - name: Cache test atomic data
-              if: ${{ !startsWith(matrix.testname, 'classicmode') }}
+              if: ${{ !startsWith(matrix.testname, 'classicmode') && !startsWith(matrix.testname, 'kilonova_2d_timedepnlte') }}
               uses: actions/cache@v6
               id: cache-testatomicdata
               with:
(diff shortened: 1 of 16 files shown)


lukeshingles:
Folded into #539 at the maintainer's request.

PR #610, opened 2026-09-17
Remove the log lines for the polarisation spectrum files

lukeshingles:
`write_specpol()` logged `Writing <name>` for `specpol.out`, `emissionpol.out`, and `absorptionpol.out`, and for their direction bin files. A different rank writes each file, and only that rank logged the name. The log of rank 0, and therefore `exspec.txt`, showed only the `emissionpol` lines. No other light curve or spectrum file has such a line.

This pull request removes the three lines. The output files and the numerical results do not change.


lukeshingles:
This change is now a commit on the branch of #609.

